PR #23, opened 2022-06-02
Add extra context for captured events

### LINKED ISSUE #16: Add extra SDK info to captured events
Currently, all the events captured by the plugin don't provide any information about the UE and plugin versions. The following values should be added to the Sentry context by default:
- Unreal Engine version
- Plugin version

Also, the plugin can react to certain common events (i.e. map loaded, game state changed etc.) and add corresponding breadcrumbs.

tustanivsky:
Resolves #16


bruno-garcia (approved):
RSLGTM but seems like we're missing tests here

PR #25, opened 2022-06-03
Fix string conversion memory leak on Android

### LINKED ISSUE #24: Memory leak due to FString-to-jstring conversion
`SentryConvertorsAndroid::StringToNative` utility function used for string conversion on Android doesn't free memory for newly created string objects. That can cause memory issues when the plugin reports many events during a short period of time.

In order to fix this `FJavaHelper::ToJavaString` function should be used instead. It wraps a plain jstring object within `FScopedJavaObject<T>` that handles memory allocation/deallocation properly.

tustanivsky:
Resolves #24

### DIFF
--- a/Source/Sentry/Private/Android/Infrastructure/SentryConvertorsAndroid.cpp
+++ b/Source/Sentry/Private/Android/Infrastructure/SentryConvertorsAndroid.cpp
@@ -59,18 +59,9 @@ jobject SentryConvertorsAndroid::SentryMessageToNative(FString message)
 	jobject messageObject = Env->NewObject(messageClass, messageCtor);
 	jmethodID setMessageMethod = Env->GetMethodID(messageClass, "setMessage", "(Ljava/lang/String;)V");
 
-	Env->CallVoidMethod(messageObject, setMessageMethod, StringToNative(message));
+	Env->CallVoidMethod(messageObject, setMessageMethod, *FJavaClassObject::GetJString(message));
 
 	return messageObject;
-}
-
-jstring SentryConvertorsAndroid::StringToNative(FString string)
-{
-	JNIEnv* Env = AndroidJavaEnv::GetJavaEnv();
-	jstring local = Env->NewStringUTF(TCHAR_TO_UTF8(*string));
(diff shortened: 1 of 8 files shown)


bruno-garcia (approved):
LGTM but lets see what @vaind says

Source/Sentry/Private/Android/Infrastructure/SentryConvertorsAndroid.cpp
```diff
@@ -59,20 +59,11 @@ jobject SentryConvertorsAndroid::SentryMessageToNative(FString message)
 	jobject messageObject = Env->NewObject(messageClass, messageCtor);
 	jmethodID setMessageMethod = Env->GetMethodID(messageClass, "setMessage", "(Ljava/lang/String;)V");
 
-	Env->CallVoidMethod(messageObject, setMessageMethod, StringToNative(message));
+	Env->CallVoidMethod(messageObject, setMessageMethod, *FJavaClassObject::GetJString(message));
```
vaind, line 62: Can you point me to API docs for `FJavaClassObject` - couldn't find it anywhere

tustanivsky: Apparently, there is no documentation for `FJavaClassObject`. You can probably check it out in UE sources ([AndroidJava.h](https://github.com/EpicGames/UnrealEngine/blob/release/Engine/Source/Runtime/Core/Public/Android/AndroidJava.h), [AndroidJava.cpp](https://github.com/EpicGames/UnrealEngine/blob/release/Engine/Source/Runtime/Core/Private/Android/AndroidJava.cpp)), though I'm not sure if that would be helpful.

vaind (approved):
LGTM TTBOMK :)

Source/Sentry/Private/Android/Infrastructure/SentryConvertorsAndroid.cpp
vaind: BTW why does the method arg need to be `(FString message)` instead of `(const FString& message)`?

PR #29, opened 2022-06-13
Create a CI/CD pipeline for plugin packaging and testing

### LINKED ISSUE #4: Create a CI/CD pipeline for plugin packaging and testing
In order to create a reproducible environment within which automated plugin packaging and testing can be performed an Unreal Engine container is required.

At the moment the most comprehensive solution out there is ue4-docker project (https://github.com/adamrehn/ue4-docker), which allows to build different types of Docker images for UE 4.20 and newer. For this particular use case the development image type is appropriate. Note that development images might take many gigabytes in size. 

Development images contain the Unreal Engine build tools and Editor, and there are strict EULA restrictions that govern their distribution ([https://unrealcontainers.com/docs/obtaining-images/eula-restrictions](https://unrealcontainers.com/docs/obtaining-images/eula-restrictions)).

Also, there are some official container images derived from ue4-docker project provided by Epic Games for UE 4.27 and newer (https://github.com/orgs/EpicGames/packages/container/package/unreal-engine)

Other considerations:
- Currently plugin doesn’t support Linux platform which can cause issues when using Linux containers
- Given images don’t support iOS and Android target platforms
- In order to package the plugin it should be placed into `Plugins/Sentry` folder of some UE project first

tustanivsky:
Resolves #4

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,29 +5,29 @@ on: [push]
 jobs:
   package-plugin-job:
     runs-on: ubuntu-latest
+    container:
+      image: ghcr.io/epicgames/unreal-engine:dev-slim-4.27
+      credentials:
+        username: ${{ secrets.UE_REPO_USER }}
+        password: ${{ secrets.UE_REPO_ACCESS_TOKEN }}
+      options: --user root
 
     steps:
-      # Checkout the UE plugin repo
       - name: Checkout
         uses: actions/checkout@v2
 
-      # Pull UE docker image and start the container
-      - name: Run Docker container
+      - name: Installing Linux Dependencies
         run: |
-          echo ${{ secrets.UE_REPO_ACCESS_TOKEN }} | docker login ghcr.io -u ${{ secrets.UE_REPO_USER }} --password-stdin
-          docker run -td --name unreal \
-            -v ${{ github.workspace }}:/ue-docker-test \
-            --workdir /ue-docker-test \
-            ghcr.io/epicgames/unreal-engine:dev-slim-4.27
+          sudo apt-get update
+          pip3 install --upgrade pip
 
-      # Install required UE dependencies
-      - name: Setup environment
-        run: |
-          docker exec unreal sudo -H apt-get update
-          docker exec unreal sudo -H pip3 install --upgrade pip
-          docker exec unreal sudo -H pip3 install ue4cli
-          docker exec unreal ue4 setroot /home/ue4/UnrealEngine
+      - name: Installing UE CLI
+        run:
+          pip3 install ue4cli
 
-      # Package plugin
+      - name: Configuring UE CLI
+        run:
+          ue4 setroot /home/ue4/UnrealEngine
+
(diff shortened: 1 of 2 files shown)


tustanivsky:
The work is still in progress, though it might be helpful to receive some early feedback on this.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,45 @@
...
+        include:
+        - unreal: ue-4-27
+          UE_IMAGE: "ghcr.io/epicgames/unreal-engine:dev-slim-4.27"
+          UE_VERSION: "ue4"
+        - unreal: ue-5-0
+          UE_IMAGE: "ghcr.io/epicgames/unreal-engine:dev-slim-5.0.2"
+          UE_VERSION: "ue5"
+
+    container:
+      image: ${{ matrix.UE_IMAGE }}
```
vaind, line 19: continuation of the change above 
```suggestion
      image: ghcr.io/epicgames/unreal-engine:dev-slim-${{ matrix.unreal }}
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,43 @@
...
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        unreal: [ue-4-27, ue-5-0]
+        include:
+        - unreal: ue-4-27
+          UE_IMAGE: "ghcr.io/epicgames/unreal-engine:dev-slim-4.27"
+        - unreal: ue-5-0
+          UE_IMAGE: "ghcr.io/epicgames/unreal-engine:dev-slim-5.0.2"
```
vaind, line 16: to simplify adding & updating in the future
```suggestion
        unreal: ['4.27', '5.0.2']
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,43 @@
...
+      options: --user root
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Installing Linux Dependencies
+        run: |
+          sudo apt-get update
+          pip3 install --upgrade pip
```
vaind, line 32: just in case we needed to install more stuff, it may be worthwhile to instead run the the container inside the GH action, as opposed to running the GH action in the container. Like this https://github.com/getsentry/sentry-unity/blob/76f0e16886974effca3f0df4aad03876b30d0d0b/.github/workflows/ci.yml#L84

vaind: Not sure the ue4cli would work with that though. also, is it worth installing just for a single call (`ue4 package`)?

tustanivsky: I've tried running the container inside the GH action in this [commit](https://github.com/getsentry/sentry-unreal/pull/29/commits/7f487849a66e7878cf566d36377445492fb23c07), though the current approach seems to look a bit cleaner to me. Not sure if that is going to make any difference atm, since both solutions work identically

vaind: sounds good

bruno-garcia (approved):

PR #35, opened 2022-07-14
Plugin dependencies update automation

### LINKED ISSUE #26: SDK names to indicate built for the Unreal package
Details in: https://github.com/getsentry/sentry-unity/issues/616

 Android: Java/Kotlin: sentry.java.android.unreal
 Android: C, C++: sentry.native.android.unreal
 iOS: Objective-C, Swift, C, C++: sentry.cocoa.unreal
 macOS: Objective-C, Swift, C, C++: sentry.cocoa.unreal
 Windows: C, C++: sentry.native.unreal

tustanivsky:
Resolves #26


Scripts/build-cocoa.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
vaind, line 1: Unless there's a reason not to, I find setting `set -euo pipefail` on the second line of every bash script helpful to avoid errors down the road (e.g. typos, silently ignored undefined variables, etc)

Scripts/build-native-linux.sh
```diff
@@ -0,0 +1,16 @@
...
+
+cmake -S "${sentryNativeRoot}" -B "${sentryNativeRoot}/build" -D SENTRY_BACKEND=breakpad -D SENTRY_SDK_NAME=sentry.native.unreal -D CMAKE_BUILD_TYPE=RelWithDebInfo
+cmake --build "${sentryNativeRoot}/build" --target sentry --parallel
+
+strip -s "${sentryNativeRoot}/build/libsentry.so" -w -K sentry_[^_]* -o "${sentryArtifactsDestination}/bin/libsentry.so"
+
+cp "${sentryNativeRoot}/build/libsentry.so" "${sentryArtifactsDestination}/bin/libsentry.dbg.so"
+cp "${sentryNativeRoot}/include/sentry.h" "${sentryArtifactsDestination}/include/sentry.h"
+
+objcopy --add-gnu-debuglink="${sentryArtifactsDestination}/bin/libsentry.dbg.so" "${sentryArtifactsDestination}/bin/libsentry.so"
```
vaind, line 16: `gnu-debuglink` should be relative path AFAIK, because the absolute path in `sentryArtifactsDestination` is only valid on the build machine.
```suggestion
objcopy --add-gnu-debuglink=libsentry.dbg.so "${sentryArtifactsDestination}/bin/libsentry.so"
```

Scripts/build-native-mac.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+export sentryNativeRoot=$1
+export sentryArtifactsDestination=$2
+
+rm -rf "${sentryArtifactsDestination}/bin/"*
+rm -rf "${sentryArtifactsDestination}/include/"*
+
+cmake -S "${sentryNativeRoot}" -B "${sentryNativeRoot}/build" -D SENTRY_BACKEND=crashpad -D SENTRY_SDK_NAME=sentry.native.unreal -DCMAKE_OSX_ARCHITECTURES="arm64;x86_64" -DCMAKE_OSX_DEPLOYMENT_TARGET=10.14
```
vaind, line 8: you may need to specify PYTHON_EXECUTABLE here (just pointing out, if it comes up in CI) - see https://github.com/getsentry/sentry-unity/pull/699/files#diff-50e91c82311ea26f2a73202525dfdf2b0a89c178ee666b2bf3ad4c84ac4c06dcR216

Scripts/checkout-submodule.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+set -euo pipefail
+
+git submodule foreach git submodule update --init --recursive
```
vaind, line 4: How is this different (in practice) from:
```suggestion
git submodule update --init --recursive
```

Scripts/update-cocoa.sh
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
```
vaind, line 1: Maybe these three scripts into one, with config being the input script, and all of the `sourcing` the common script. Removes duplication a bit. See for example https://github.com/getsentry/sentry-react-native/blob/main/scripts/update-wizard.sh that's the "updater.yml" target script and itself includes a common `update-package-json.sh`

vaind (approved):
Looks pretty good

Scripts/build-native-linux.sh
tustanivsky: Good catch!

Scripts/build-cocoa.sh
tustanivsky: Added `set -euo pipefail` for all scripts where it was missing.

Scripts/update-cocoa.sh
tustanivsky: I've raised a separate issue for this

PR #45, opened 2022-07-28
Automatic debug symbols upload for Android

### LINKED ISSUE #14: Add automatic debug symbols upload for Android
UPL ([Unreal Plugin Language](https://docs.unrealengine.com/en-US/SharingAndReleasing/Mobile/UnrealPluginLanguage/)) configuration file allows enhancing build.gradle which can be used for automatic debug symbols upload with [Sentry Android Gradle Plugin](https://docs.sentry.io/platforms/android/gradle/).

tustanivsky:
Resolves #14

### DIFF
--- a/Source/Sentry/Sentry_Android_UPL.xml
+++ b/Source/Sentry/Sentry_Android_UPL.xml
@@ -37,7 +37,9 @@
         <insert>
             -dontwarn com.sentry.unreal.**
             -keep class com.sentry.unreal.** { *; }
+            -keep class io.sentry.** { *; }
             -keep interface com.sentry.unreal.** { *; }
+            -keep interface io.sentry.** { *; }
         </insert>
     </proguardAdditions>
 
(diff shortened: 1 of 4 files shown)


Source/Sentry/Sentry_Android_UPL.xml
```diff
@@ -37,10 +43,27 @@
...
 
+    <buildGradleAdditions>
+        <if condition="bUploadSymbols">
+            <true>
+                <insert>
+                    apply plugin: 'io.sentry.android.gradle'
+
+                    sentry {
+                        uploadNativeSymbols = true
+                        includeNativeSources = true
```
bitsandfoxes, line 60: I might be reading this wrong but does that mean it's enabled by default?

tustanivsky: This entry will end up in a resulting `build.gradle` only if user enabled automatic debug symbols upload in plugin settings (see the `bUploadSymbols` check right above)

bitsandfoxes: Ah, gotcha, so this is just the template. Thanks!

PR #52, opened 2022-08-04
Fixing broken link

Edstub207:


Edstub207:
@tustanivsky FYI

tustanivsky:
Good catch! Thanks for pointing that out

PR #53, opened 2022-08-04
Clarify which endpoint is required

Edstub207:
Ensuring it's clear it needs to be the Unreal Engine endpoint - Not just the standard DSN.


Edstub207:
@tustanivsky FYI

Edstub207:
Invalid info

PR #59, opened 2022-08-11
Add debug symbols upload

### LINKED ISSUE #47: Automatic debug symbols upload for iOS
UE allows adding the `PostBuildSteps` section in .uplugin file where a separate shell script for symbols upload can be called. In case with iOS at this stage .dSYM file should be already available in `<ProjectRoot>/Binaries/IOS`. Also, `sentry-cli` seems to be a perfect tool for handling the upload process etc.

### LINKED ISSUE #57: Unreal Engine Debug File Uploads
I'm not sure if this is possible, if it is, it might be a while away, but is it possible to have the functionality to automatically upload the debug symbols (https://docs.sentry.io/product/cli/dif/) based on what Sentry requires/expects, possibly as a "plug and play" commandlet or build graph script.

As while it's easy to do, having the functionality out of the box, with different options, such as stripping symbols and uploading for each config etc would be pretty handy at the click of a button, especially if I've built code locally and want to upload for whatever reason.  

Especially when console support becomes a thing, this will be useful for anyone using the plugin - Instead of everyone having to write a custom solution.

### LINKED ISSUE #60: Automatic Sentry CLI update in CI
In order to automate debug symbols upload for certain platforms Sentry CLI was added to the plugin. To keep this tool up to date existing CI workflow for plugin dependencies can be extended.

tustanivsky:
Resolves #47
Resolves #57 
Resolves #60

It will also make sense to automate Sentry CLI update in a way similar to other plugin dependencies (#60)


plugin-dev/Scripts/upload-debug-symbols-win.ps1
```diff
@@ -0,0 +1,89 @@
...
+            }  
+            $name,$value = $matches[1..2]  
+            $ini[$section][$name] = $value  
+        }  
+    }  
+
+    return $ini  
+}
+
+$ConfigIni = ParseIniFile "$ConfigPath\DefaultEngine.ini"
```
vaind, line 68: Is this file normally under version control? If so, we shouldn't use it for sentry CLI configuration (a.k.a `sentry.properties`) as that's supposed to be excluded from git due to the auth-token.

vaind (changes requested):

plugin-dev/Scripts/upload-debug-symbols-win.ps1
tustanivsky: Yes, `DefaultEngine.ini` is an essential part of any UE project that keeps all of its settings (including plugins) and usually it's under source control. Though I don't think it can cause any issues for plugin users since they are likely to take care of securing their project config files. Or did I get it wrong?

vaind: Even if it's in a private repo, unless it's a single-person project, it's best not to commit auth tokens into source control. Additionally, this disqualifies any open-source games.

vaind: I'd suggest sticking to a single sentry.propertiies file, regardless of the platform

tustanivsky: Don't we have the exact same issue with Unity then? AFAIK there auth token is stored in a separate asset (`SentryCliOptions` scriptible object) which apparently could be a part of the game as well

vaind: It's not part of the game and it's recommended to exclude it from source control as well..

tustanivsky: > I'd suggest sticking to a single `sentry.propertiies` file, regardless of the platform

We have a template for `sentry.propertiies` file in the repo which is filled with actual user credentials (set in DefaultEngine.ini) when it comes to uploading debug symbols.

Do you mean we should use it both for Windows and other platforms? I recall having some issues with taking this approach on Windows, but I'll double check

vaind: yes, we use the sentry.properties file on windows without an issue in Unity

plugin-dev/Sentry.uplugin
```diff
@@ -21,5 +21,17 @@
...
+			"sh $(PluginDir)/Scripts/upload-debug-symbols.sh $(TargetPlatform) $(TargetName) $(TargetType) $(ProjectDir) $(PluginDir)"
+		],
+		"Linux": [
+			"sh $(PluginDir)/Scripts/upload-debug-symbols.sh $(TargetPlatform) $(TargetName) $(TargetType) $(ProjectDir) $(PluginDir)"
+		],
+		"Win64": [
+			"powershell -ExecutionPolicy RemoteSigned -File $(PluginDir)/Scripts/upload-debug-symbols-win.ps1 $(TargetPlatform) $(TargetName) $(TargetType) $(ProjectDir) $(PluginDir)"
+		]
+	}
 }
```
vaind, line 37: Can we skip the step of merging these binaries to the repo, thus increasing `.git` directory size forever and instead add the download to CLI https://github.com/getsentry/sentry-unity/blob/938bb4787911171c443b499642e76c9780b13e3c/scripts/download-sentry-cli.ps1

tustanivsky: So instead of keeping Sentry CLI executable in the repo we'll be downloading it every time PostBuildSteps are triggered, right?

vaind: No, we download it in ci and package in the release. Same as other ThirdParty libs

vaind: search for that script name in the sentry-unity repo to see how it's used in CI

tustanivsky: Oh, I've misunderstood the original question. Yes, that totally makes sense

plugin-dev/Scripts/upload-debug-symbols-win.ps1
vaind: Practical illustration we shouldn't store the settings in the same settings file - from the in-repo sample PR (#64):

image

tustanivsky: > How can we move that to a different location so the "shared" file doesn't change?

We can replace user credentials here with a path to `sentry.properties` file which won't be publicly accessible. Later we can add some editor tool that will automate the creation of this file. What do you think?

vaind: sounds good to me

tustanivsky: Alright, no more user credentials in `DefaultEngine.ini`. Now in order to obtain auth token and other stuff for Sentry CLI path to `sentry.properties` file has to be provided firs. Apparently not an ideal solution since it may vary on different machines, though at least we've removed the risk of compromising sensitive data

vaind:
@tustanivsky for Android, we need to update the UPL file - how can we reference the ProjectDir and the Sentry.PropertiesFilePath here? https://github.com/getsentry/sentry-unreal/blob/3a0d8a143a8140b8de2b3b4b615ce70a26552fc4/plugin-dev/Source/Sentry/Sentry_Android_UPL.xml#L15

tustanivsky:
> @tustanivsky for Android, we need to update the UPL file - how can we reference the ProjectDir and the Sentry.PropertiesFilePath here?
> 
> https://github.com/getsentry/sentry-unreal/blob/3a0d8a143a8140b8de2b3b4b615ce70a26552fc4/plugin-dev/Source/Sentry/Sentry_Android_UPL.xml#L15

`$S(ProjectDir)` should give us the path to the project directory and the `sentry.properties` file can be referenced via `$S(PropertiesFilePath)` variable which is defined above.

It actually looks like something went wrong during recent merges - `src` should be `$S(PropertiesFilePath)`

vaind:
> It actually looks like something went wrong during recent merges - `src` should be `$S(PropertiesFilePath)`

fixed

please have a look at my changes and feel free to merge.

PR #61, opened 2022-08-15
Refactor: repo structure

### LINKED ISSUE #27: Project structure
I've had a brief look at this project and tried following the CONTRIBUTING.md to build the project and I have a few suggestions.

Currently, the project is supposed to be checked out into an existing Unreal Engine project as a plugin. This means a developer of sentry-unreal SDK must first create a new UE project (sample), to be able to do any development. This increases the cost-of-entry for new SDK developers. Instead, I'd suggest having a look at https://github.com/getsentry/sentry-unity and think about the following changes:

* [x] move the existing contents to a new directory, e.g. what package-dev is in the Unity SDK, maybe `plugin` or `plugin-dev` is a good name? - #61 
* [x] move dependency builds to separate CI job & allow downloading for plugin developers - thus moving the binaries from the repository - #63
* [x] add script for packaging and release - #62 
* [x] add a sample project that already includes the development version of the plugin - #64 
* [ ] add a test project that includes a plugin that's ready for distribution (see also #4), which may be different than the dev version - #77

vaind:
1st step of #27


plugin-dev/Sentry.uplugin
```diff
@@ -14,7 +14,7 @@
 	"IsBetaVersion": false,
 	"IsExperimentalVersion": false,
 	"Installed": false,
-	"Modules": [
+	"modules": [
```
tustanivsky, line 17: I believe we should leave `Modules` here since it is a pre-determined field for the plugin descriptor file (https://docs.unrealengine.com/5.0/en-US/API/Runtime/Projects/FPluginDescriptor/Modules/)

plugin-dev/Source/Sentry/Private/SentryModule.cpp
```diff
@@ -7,7 +7,7 @@
 #include "UObject/Package.h"
 #include "UObject/UObjectGlobals.h"
 #include "Interfaces/IPluginManager.h"
-#include "Modules/ModuleManager.h"
+#include "modules/ModuleManager.h"
```
tustanivsky, line 10: Let's keep the included directives without changes (probably an auto replacement issue here)

plugin-dev/Source/Sentry/Public/SentryModule.h
```diff
@@ -3,7 +3,7 @@
 #pragma once
 
 #include "CoreMinimal.h"
-#include "Modules/ModuleInterface.h"
+#include "modules/ModuleInterface.h"
```
tustanivsky, line 6: Same here

plugin-dev/Sentry.uplugin
vaind: except these are not "Modules" in Unreal world? https://docs.unrealengine.com/4.27/en-US/ProgrammingAndScripting/ProgrammingWithCPP/Modules/

tustanivsky: These are exactly the same "Modules". Probably my previous comment was a bit misleading - I've meant we should leave "Modules" instead of "modules" (all lowercase) to keep things consistent here

vaind: my bad, I was referring to the repo/modules with gitmodules. this line is a mistake of course

```suggestion
	"Modules": [
```

plugin-dev/Source/Sentry/Private/SentryModule.cpp
vaind:
```suggestion
#include "Modules/ModuleManager.h"
```

plugin-dev/Source/Sentry/Public/SentryModule.h
vaind:
```suggestion
#include "Modules/ModuleInterface.h"
```

tustanivsky (approved):

PR #62, opened 2022-08-15
Chore/release ci

vaind:
2nd step of #27  - keeping separately from SDK build scripts updates so that this doesn't blow up too much and is (much) easier to review.

The current plan is to stick to single repo and publish the three packages (one for each EngineVersion) as zip files as  GH release artifacts.


tustanivsky:
In order to publish a new plugin at UE Marketplace we need to support three latest engine versiona (currently those are 4.26.0, 4.27.0 and 5.0.0). Since Sentry plugin belongs to `Code Plugins` category supported engine version should be specified explicitly in `Sentry.uplugin` file according to this guideline (https://www.unrealengine.com/en-US/marketplace-guidelines?sessionInvalidated=true#261a). Is it possible to create three separate packages with the right engine version there when making a release on GitHub?

vaind:
> In order to publish a new plugin at UE Marketplace we need to support three latest engine versiona (currently those are 4.26.0, 4.27.0 and 5.0.0). Since Sentry plugin belongs to `Code Plugins` category supported engine version should be specified explicitly in `Sentry.uplugin` file according to this guideline (https://www.unrealengine.com/en-US/marketplace-guidelines?sessionInvalidated=true#261a). Is it possible to create three separate packages with the right engine version there when making a release on GitHub?

It is possible, but where do we push those? To three separate repositories? Is this the only way? Maybe we can specify the minimum supported version instead? How do others plugins handle this?

tustanivsky:
> > In order to publish a new plugin at UE Marketplace we need to support three latest engine versiona (currently those are 4.26.0, 4.27.0 and 5.0.0). Since Sentry plugin belongs to `Code Plugins` category supported engine version should be specified explicitly in `Sentry.uplugin` file according to this guideline (https://www.unrealengine.com/en-US/marketplace-guidelines?sessionInvalidated=true#261a). Is it possible to create three separate packages with the right engine version there when making a release on GitHub?
> 
> It is possible, but where do we push those? To three separate repositories? Is this the only way? Maybe we can specify the minimum supported version instead? How do others plugins handle this?

I was thinking about adding three separate archives for each engine version with a properly modified .uplugin file to the release and leaving the repo source code itself without any changes. The idea is pretty similar to other Sentry repos (i.e. for Sentry CLI https://github.com/getsentry/sentry-cli/releases/tag/2.5.0 executables for different platforms attached to the release). Also, links to such archives can be used for plugin update submission at UE Marketplace.

Unfortunately providing only the minimum supported version is not an option and the approach suggest above seems to be a common practice.

Alternatively repo can contain three subfolders (i.e. Sentry_4.26, Sentry_4.27 and Sentry_5.0) for engine-specific plugin version each, however that looks like an overkill since there will be difference only in one file.

vaind:
OK, thanks for clarifying. Having the output repo structured to include multiple distributions sounds bad but also like the right think to do 🤕 

What do we actually upload to UE marketplace? The result of `docker exec unreal ue4 package`? Also related to my question here: https://github.com/getsentry/sentry-unreal/pull/62/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR49

tustanivsky:
UE Marketplace has its own dashboard for publishers where one has to provide links to downloadable archives with plugin source code for each engine version that is supported:

The_business_of_building_code_plugins_for_game_engines___by_Taras_Leskiv___nineva___Medium

That's what makes me think that having those at repo release page will be a convenient thing for submitting updates.

Regarding the existing plugin packaging workflow - in our case this is more like a validation that there are no build errors. Epic Games does exactly the same thing after plugin update submission anyway and takes care of deploying a pre-build version to the marketplace

.gitignore
```diff
@@ -32,3 +32,8 @@ Intermediate/*
 
 # Other
 .DS_Store
```
tustanivsky, line 34: Can you also change `Binaries/*` and `Intermediate/*` .gitignore entries to `plugin-dev/Binaries/*` and `plugin-dev/Intermediate/*` respectively please? I've missed that part for the previous PR

tustanivsky (approved):

.craft.yml
```diff
@@ -0,0 +1,14 @@
+minVersion: 0.24.0
+changelogPolicy: auto
+preReleaseCommand: pwsh scripts/bump-version.ps1
+targets:
```
bitsandfoxes, line 4: The is the "UPM" following in another PR?

bitsandfoxes (approved):

.craft.yml
bitsandfoxes: > The current plan is to stick to single repo and publish the three packages (one for each EngineVersion) as zip files as GH release artifacts.

I see.

vaind: no more UPM, see PR description - for now we're sticking to a single repo

PR #63, opened 2022-08-17
Remove dependency binaries

### LINKED ISSUE #28: Avoid precompiled binaries in the repo
It would be nice if we could avoid having binaries pushed in the repo, such as Android aar/jar files, crashpad_handler executables, etc. Ideally, the files should be downloaded from their repositories with a script that can be run both locally & in CI before packaging (see #4) and check a hard-coded checksum. Updating to a newer dependency would be just updating the tag & checksum in this script. 

Alternatively, we could run the build in CI in this repo, as we do for sentry-unity, but that complicates things here a bit and is slower so unless there's strong incentive to do that, having a script should be enough IMO.

### LINKED ISSUE #41: Refactor build scripts for plugin dependencies
There is some code duplication in scripts that are used for updating plugin dependencies.

"Maybe these three scripts into one, with config being the input script, and all of the `sourcing` the common script. Removes duplication a bit. See for example https://github.com/getsentry/sentry-react-native/blob/main/scripts/update-wizard.sh that's the "updater.yml" target script and itself includes a common `update-package-json.sh`"

_Originally posted by @vaind in https://github.com/getsentry/sentry-unreal/pull/35#discussion_r923059130_

vaind:
* closes #28
* closes #41

3rd step of #27


tustanivsky:
We should call `scripts/download-sdks.sh` on the client in order to download the latest dependencies, right? If so, probably  some kind of pre-processing in the plugin to trigger that (i.e. `PreBuildSteps` in `Sentry.uplugin` file) will be required. Ofc calling it manually is an option as well

vaind:
> We should call `scripts/download-sdks.sh` on the client in order to download the latest dependencies, right? If so, probably some kind of pre-processing in the plugin to trigger that (i.e. `PreBuildSteps` in `Sentry.uplugin` file) will be required. Ofc calling it manually is an option as well

I wouldn't add it as a pre-processing because it's only for sentry-unreal developers, not plugin end-users

scripts/download-sdks.sh
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+cd "$(dirname $0)/../plugin-dev/Source/ThirdParty"
```
tustanivsky, line 4: We have to make sure that the `ThirdParty` directory exists first. Apparently, after removing all those binaries it has gone too

vaind:
```suggestion
mkdir -p "$(dirname $0)/../plugin-dev/Source/ThirdParty"
cd "$(dirname $0)/../plugin-dev/Source/ThirdParty"
```

PR #68, opened 2022-08-19
Revert Android libraries importing mechanism

### LINKED ISSUE #42: Android SDK integration doesn't use locally build package
Crashes captured on Android should provide info regarding the native SDK name which is used under the hood by the UE plugin (`sentry.native.android.unreal`), however default value is used instead.

#26 

SDK name for sentry-java is specified in a corresponding build script: https://github.com/getsentry/sentry-unreal/blob/main/Scripts/build-java.sh

tustanivsky:
Closes #42

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@
 - Bump Java SDK (Android) to v6.3.1 ([#51](https://github.com/getsentry/sentry-unreal/pull/51))
   - [changelog](https://github.com/getsentry/sentry-java/blob/main/CHANGELOG.md#631)
   - [diff](https://github.com/getsentry/sentry-java/compare/6.3.0...6.3.1)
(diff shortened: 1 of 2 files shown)


vaind (approved):

vaind:
For future reference - the reason is that `AARImports` doesn't provide much options, basically just adding a standard `implementation` dependency or a `maven` repository. The docs are non-existent but the we can see this in the [source code](https://github.com/EpicGames/UnrealEngine/blob/d11782b9046e9d0b130309591e4efc57f4b8b037/Engine/Source/Programs/UnrealBuildTool/Platform/Android/UEDeployAndroid.cs#L5428)

The problem is twofold. The referenced repository actually didn't have the required maven structure; and even if it did, we couldn't force gradle to **not** use any other matching dependency from a previous repo (e.g. maven central). The only way this could be done at the moment would be rename the packages so they're unique to the unreal project. This would be more work with no benefit as opposed to rolling back to copying AARs so the decision was to abandon AARImports until an issue with the restored approach is found.

PR #81, opened 2022-08-23
ci: build sample app; fix packaging

vaind:
originally, this was about adding a sample app build, but I've had to make other changes in the process
1. having `--user root` when running the docker container hid the issue with the package permissions being broken
2. powershell `compress-archive` doesn't create directories - this changes the snapshot file
3. I'm not 100 % sure about the cause, but there were permission issues with the extracted package/Source/ThirdParty directory having 0 permissions on macOS/Linux - may be either `compress-archive`, or the directories being created by GH download-artifact action, or the combination of the two. In any case, changing to `zip` fixes the issue.


tustanivsky (approved):

scripts/packaging/package.snapshot
```diff
@@ -166,24 +199,30 @@ Source/ThirdParty/IOS/Sentry.framework/PrivateHeaders/SentryLog.h
...
 Source/ThirdParty/Win64/bin/crashpad_handler.exe
 Source/ThirdParty/Win64/bin/sentry.dll
 Source/ThirdParty/Win64/bin/sentry.pdb
+Source/ThirdParty/Win64/include/
 Source/ThirdParty/Win64/include/sentry.h
+Source/ThirdParty/Win64/lib/
 Source/ThirdParty/Win64/lib/sentry.lib
-Source/ThirdParty/CLI/sentry-cli-Darwin-universal
-Source/ThirdParty/CLI/sentry-cli-Linux-x86_64
-Source/ThirdParty/CLI/sentry-cli-Windows-x86_64.exe
```
vaind, line 189: FYI @tustanivsky - when you need to make changes to this file, run the following locally. 
```shell-script
pwsh ./scripts/packaging/pack.ps1
pwsh ./scripts/packaging/test-contents.ps1 accept
```
(it sorts the output correctly so in this case the manual changes got sorted in this PR even though they don't belong here)

PR #86, opened 2022-08-26
Suppressing warnings (Needed for when using Warnings as Errors) 

### LINKED ISSUE #85: Code fails to compile when using Warnings As Errors on the target
### Environment

How do you use Sentry?
Sentry.io 

Which version of the SDK?
Source code from git with libs manually integrated. 

Which version of Unreal?

5.0.0.

### Steps to Reproduce

1. Enable 	bWarningsAsErrors = true; in the target
2. Compile 

### Expected Result
Compile completes. 

### Actual Result

Compile fails: 
warning C6011: Dereferencing NULL pointer 'ptr'. : Lines: 92, 93

Edstub207:
Closes #85


tustanivsky (approved):

Edstub207:
@tustanivsky I'm not sure what to do with this failure - Is it okay if I leave it with you?

PR #125, opened 2022-10-04
Fix incomplete type forward declaration

### LINKED ISSUE #124: Plugin packaging error in consoles build
```
In file included from D:/workspace/ProjectBuild/Project/Intermediate/Build/PS5/Project/Development/Sentry/Module.Sentry.cpp:22:
D:\workspace\ProjectBuild\Project\Plugins\Sentry\Source\Sentry\Private\SentrySubsystem.cpp(65,21): error: member access into incomplete type 'ISentrySubsystem'.
SubsystemNativeImpl->InitWithSettings(Settings);
D:\workspace\ProjectBuild\Project\Plugins\Sentry\Source\Sentry\Public\SentrySubsystem.h(19,7): note: forward declaration of 'ISentrySubsystem'.
class ISentrySubsystem;
```
I have found that the above error occurs when building to PS5 target with UnrealEngine5.
This seems to happen even if I set Sentry to "Enabled: false" in .uproject or WhitelistPlatforms in .uplugin.
```
#include "Interface/SentrySubsystemInterface.h"
```
It seems I need to add this to SentrySubsystem.cpp for the build to succeed. It would be great if you could fix this.

tustanivsky:
Closes #124

### DIFF
--- a/plugin-dev/Source/Sentry/Private/SentrySubsystem.cpp
+++ b/plugin-dev/Source/Sentry/Private/SentrySubsystem.cpp
@@ -12,15 +12,13 @@
 #include "Misc/EngineVersion.h"
 #include "Misc/CoreDelegates.h"
 
+#include "Interface/SentrySubsystemInterface.h"
+
 #if PLATFORM_ANDROID
 #include "Android/SentrySubsystemAndroid.h"
-#endif
-
-#if PLATFORM_IOS
+#elif PLATFORM_IOS
 #include "IOS/SentrySubsystemIOS.h"
-#endif
-
-#if PLATFORM_WINDOWS || PLATFORM_MAC || PLATFORM_LINUX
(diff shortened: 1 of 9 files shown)


plugin-dev/Source/Sentry/Private/SentrySubsystem.cpp
```diff
@@ -12,15 +12,13 @@
 #include "Misc/EngineVersion.h"
 #include "Misc/CoreDelegates.h"
 
+#include "Interface/SentrySubsystemInterface.h"
```
tustanivsky, line 15: Adding this should be enough to resolve the missing forward declaration issue. The same approach is used for some entities that are supported only on mobile platforms and that doesn't break the compilation on desktop.

bitsandfoxes (approved):

PR #152, opened 2022-11-11
Add editor menu for automatic CRC configuration

### LINKED ISSUE #58: Automatic CRC configuration
UE allows adding the PostBuildSteps section in .uplugin file where a separate shell script for CRC configuration can be called. This should allow users to skip steps described in [this guide](https://docs.sentry.io/platforms/unreal/setup-crashreport/). It's also worth considering how the CRC setup should work (via local/global config file).

tustanivsky:
Now there is a button in the plugin settings menu that partially automates CRC configuration process so that users won't need to [change `DefaultEngine.ini` in the engine dir manually](https://docs.sentry.io/platforms/unreal/setup-crashreporter/#configure-the-crash-reporter-endpoint). Since this change affects global engine settings another button that allows reverting it was added as well.

The approach with adding PostBuildStep suggested earlier in the linked issue would be less user-friendly and flexible compared to the custom editor UI.

Closes #58

### DIFF
--- a/plugin-dev/Source/Sentry/Private/SentrySettings.cpp
+++ b/plugin-dev/Source/Sentry/Private/SentrySettings.cpp
@@ -13,6 +13,7 @@ USentrySettings::USentrySettings(const FObjectInitializer& ObjectInitializer)
 {
 	DsnUrl = TEXT("");
 	Release = TEXT("");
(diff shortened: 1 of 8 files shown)


tustanivsky:
image

bitsandfoxes (approved):
It's really really awesome!

Issue #44, opened 2022-07-27
Is automatic exception catching and reporting functionnal ?


SBRK:
### Environment

Sentry SaaS (sentry.io) 
Plugin 0.0.1
Unreal 5.0.3
Windows 10

### Steps to Reproduce

1. Configure Sentry plugin and Sentry project
2. Capture a custom "Hello World" message => works ok and is visible in Sentry issues panel
3. Trigger an exception in Unreal

### Expected Result
Error gets reported in Sentry issues

### Actual Result
Nothing gets sent

The documentation of the plugin insinuate that it should automatically send crashes. It doesn't seem to be the case. Is there anything special to do to make it work ? Is it simply not supported yet ?

tustanivsky:
Hey @SBRK , in order to capture crashes on Windows please [configure Crash Reporter](https://docs.sentry.io/platforms/unreal/setup-crashreport/) first.
This step won't be required once the next engine version goes live, however for now UE has some limitations in terms of exception handling on desktop.

SBRK:
[Crash Reporter seems to be broken for me at the moment](https://github.com/getsentry/sentry/issues/37106) so I was looking at this new plugin as an alternative

tustanivsky:
@Edstub207 
1. For Sentry plugin configuration (`ProjectSettings` -> `Plugins` -> `Sentry`) you should use the DSN. However, for the embedded Crash Reporter UE specific version has to be provided as it is described [here](https://docs.sentry.io/platforms/unreal/setup-crashreport/#configure-the-crash-reporter-endpoint)
2. No, currently there is no support for that

bruno-garcia:
We heard this will be soon possible in some 5.1 or future version of Unreal.

vaind:
I can't find any info on when is 5.1 expected to be released, though.

In any case, this looks to me like a documentation issue at this time - we should clearly indicate in the Configure section https://docs.sentry.io/platforms/unreal/#configure that there is another step that should be taken.

Edstub207:
@vaind Epic have branched for 5.1 already, so it should be in the next few weeks/months.

tustanivsky:
[...]

PR #175, opened 2022-12-12
Automatic crash capturing for UE 5.1 (Windows only)

tustanivsky:
Related to #44


tustanivsky:
Currently, automatic crash capturing works only in the editor due to Unreal's bug. This issue should be addressed with this PR (https://github.com/EpicGames/UnrealEngine/pull/9931). Alternatively, if one is using the engine built from source corresponding change can be applied in order to enable automatic crash capturing in packaged game build.

PR #204, opened 2023-02-02
feat: expose environment setting

Zaratusa:
The environment settings of the SDKs were not exposed, but they can be very useful to filter out events and crashes from test environments for example.


tustanivsky:
@Zaratusa Looks good! Thank you for submitting this PR

Issue #188, opened 2023-01-05
Libsentry.so is not resolvable warning


Edstub207:
### Environment
Unreal 

How do you use Sentry?
Sentry SaaS (sentry.io)

Which version of the SDK?
Latest from Main 

How did you install the package? (Git-URL, Assetstore)
Manually

Which version of Unreal?
5.1

Is this happening in Unreal (editor) or on a player like Android, iOS, Windows?
Compiling the server for Linux 

### Steps to Reproduce
1. Compile the server for a project Linux against the Linux platform on a Windows Build agent with warnings as errors and strict compile enabled 

### Expected Result
No warnings or errors

### Actual Result
Library 'E:\Horde\Work\Sync\Project\Plugins\External\SentryUnrealPlugin\Binaries\Linux\libsentry.so' was not resolvable to a file when used in Module 'Sentry', assuming it is a filename and will search library paths for it. This is slow and dependency checking will not work for it. Please update reference to be fully qualified alternatively use PublicSystemLibraryPaths if you do intended to use this slow path to suppress this warning.

tustanivsky:
This issue was addressed in the scope of #190

tustanivsky:
Just a quick update regarding this warning - in order to fix it I had to put back the code for copying plugin dependencies to `Binaries` folder before calling the `PublicAdditionalLibraries.Add("...\libsentry.so")` method in `Sentry.Build.cs`. The thing is that even though the `RuntimeDependencies.Add(dst, src)` copies the binaries on its own it happens at some later build stages and by the time of "adding" `libsentry.so` they aren't there yet. Linking [this](https://github.com/getsentry/sentry-unreal/commit/dc7db942868dae260200657f6b718d2f40ed22a9#commitcomment-94273387) conversation for extra context

PR #206, opened 2023-02-03
Fixing Staging problem

### LINKED ISSUE #205: Unable to copy CrashPad_Handler.exe when staging on a different machine
### Environment

How do you use Sentry?
Sentry.io

Which version of the SDK?
Latest from Main

Git

Which version of Unreal?
5.1

Build Farm

### Steps to Reproduce

1. Compile Editor
2. Compile Client
3. Cook (On a different agent, using the artifacts from the first jobs)
4: Attempt to stage (On the same agent as the cook) 

### Expected Result

No errors

### Actual Result
Skip copying file E:\Horde\Work\Inc\Sync\Redacted\Plugins\External\SentryUnrealPlugin\Binaries\Win64\crashpad_handler.exe because it doesn't exist.
ERROR: Failed to copy E:\Horde\Work\Inc\Sync\Redacted\Plugins\External\SentryUnrealPlugin\Binaries\Win64\crashpad_handler.exe to E:\Horde\Work\Inc\Sync\Redacted\Saved\StagedBuilds\WindowsClient\Trident\Plugins\External\SentryUnrealPlugin\Binaries\Win64\crashpad_handler.exe

Edstub207:
Closes #205

### DIFF
--- a/plugin-dev/Source/Sentry/Sentry.Build.cs
+++ b/plugin-dev/Source/Sentry/Sentry.Build.cs
@@ -157,7 +157,6 @@ public class Sentry : ModuleRules
 		string BinariesSymbolsPath = Path.Combine(BinariesPath, libname + DebugSymbolsExtension);
 
 		CopyPluginBinary(SourceDynamicLibPath, BinariesDynamicLibPath, BinariesPath);
-		RuntimeDependencies.Add(BinariesDynamicLibPath);
 
 		if (Target.Platform == UnrealTargetPlatform.Win64)
 		{
@@ -165,7 +164,6 @@ public class Sentry : ModuleRules
 			PublicDelayLoadDLLs.Add(libname + DynamicLibExtension);
 
 			CopyPluginBinary(SourceSymbolsPath, BinariesSymbolsPath, BinariesPath);
-			RuntimeDependencies.Add(BinariesSymbolsPath);
 		}
 		if (Target.Platform == UnrealTargetPlatform.Linux)
 		{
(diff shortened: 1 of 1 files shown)


Edstub207:
@tustanivsky I think I've hit this problem in other places as well, I'm not sure if you want to do this in a better way?

tustanivsky:
> @tustanivsky I think I've hit this problem in other places as well, I'm not sure if you want to do this in a better way?

Don't we need to make similar changes for `sentry.dll` and `sentry.pdb` as well?

Also, I was wondering why the `crashpad_hadler.exe` couldn't be found if it is copied to the plugin's `Binaries` folder right before  adding the corresponding runtime dependency

tustanivsky:
I recall we've switched to using `RuntimeDependencies.Add(Dest, Src)` instead of copying plugin dependencies manually a while ago however it turned out that this approach wasn't working properly so I've rolled back these changes. This [conversation](https://github.com/getsentry/sentry-unreal/issues/188#issuecomment-1383613427) should provide extra context

Edstub207:
> I recall we've switched to using `RuntimeDependencies.Add(Dest, Src)` instead of copying plugin dependencies manually a while ago however it turned out that this approach wasn't working properly so I've rolled back these changes. This [conversation](https://github.com/getsentry/sentry-unreal/issues/188#issuecomment-1383613427) should provide extra context

👍 I have verified the latest version of this change works, when staging WindowsServer, WindowsClient and Linux Server.

tustanivsky:
A similar fix can be found in #211. The only difference there is that `RuntimeDependencies.Add` calls stayed at their old place in order to keep binaries copying and actual dependencies management separate. Also, I had some concerns about whether it is safe to add runtime dependency only if the corresponding binary wasn't copied to the `Binaries` folder earlier.

PR #208, opened 2023-02-03
feat: automatically read release & environment

Zaratusa:
The release and environment can automatically be read in the constructor and still be overridden in the config files. This should be good basic setup for most project.


bitsandfoxes:
I like where this is going!

tustanivsky:
Looks good!

PR #246, opened 2023-04-01
Update Android layer implementation

tustanivsky:
This PR is supposed to:

- Prevent potential memory leaks caused by incorrect `jobject` references management
- Ease understanding and maintenance of the code which relies on JNI

As a result `SentryJavaObjectWrapper` class was introduced which provides a unified API for calling both instance and static native Java methods from C++. Currently it has two most common use-cases:

1. Serve as a base class for Sentry/JVM native class wrappers
2. Utility for calling static methods from `SentryBridgeJava` class

`SentryJavaObjectWrapper` ensures that the global reference to owned `jobject` is released upon its destruction.

Also, `CallObjectMethod` returns `FScopedJavaObject<T>` instead of raw `jobject` reference in order to properly release its local reference while the caller is responsible to decide whether it has to be added to global scope.


bitsandfoxes (approved):
🔥

PR #249, opened 2023-04-05
Build: Fix handling of third party dependencies

VesCodes:
This generally simplifies the complex manual handling of dependency files and resolves some teething problems we hit against in a distributed build system.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Fix snapshot update script ([#253](https://github.com/getsentry/sentry-unreal/pull/253))
 - Fix debug symbol uploading scripts ([#261](https://github.com/getsentry/sentry-unreal/pull/261))
+- Fix handling of third party dependencies ([#249](https://github.com/getsentry/sentry-unreal/pull/249))
 
 ### Dependencies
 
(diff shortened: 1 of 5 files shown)


Edstub207:
@VesCodes Has this been tested with staging and packaging builds?

IIRC  there have been a few attempts at doing similar reworks before and we've had to add the Copy step also I noticed the step which makes the files writeable for UGS binaries was removed too!

tustanivsky:
> This generally simplifies the complex manual handling of dependency files and resolves some teething problems we hit against in a distributed build system.

I've noticed that with the suggested changes `crashpad_handler` copying has been removed as well. It's basically an essential part of the plugin required for proper crash reporting unless you're using some other backend for `sentry-native`.

VesCodes:
@tustanivsky the crashpad handler is marked as a runtime dependency which is copied to the same directory as the target executable by UBT. The expression on line 89 does a wildcard (notice the `...`) copy of all files under `Source/ThirdParty/<Platform>/bin`. I'll follow up with a small change to make this more explicit as we did end up doing that to separate the debug files that also reside under the `bin` directory for some platforms.

@Edstub207 I found the existing code requires some additional work so that these dependencies are staged. This removes that requirement as they are now properly tagged as output from a target's compilation process and appropriately tracked alongside its products.

I invite you all to try this of course :)

Edstub207:
@VesCodes Okay, so far it seems okay in testing in terms of staging files based on testing. 

I suspect we'll encounter the aforementioned ReadOnly file problem due to the following being removed, but not in a place where I can check the Build.cs changes in to reproduce this issue. 
```
// Make binary writeable to avoid issues with UGS Binary Zips during sync (Perforce is usually read-only by default)
File.SetAttributes(DestFile, File.GetAttributes(DestFile) & ~FileAttributes.ReadOnly);
```

VesCodes:
> I suspect we'll encounter the aforementioned ReadOnly file problem due to the following being removed, but not in a place where I can check the Build.cs changes in to confirm this issue.

@Edstub207 that gets handled by UAT during staging.

VesCodes:
Noticed an issue with the existing pathing to DLLs loaded at runtime - we didn't catch them since we don't test on Mac and that's seemingly the only platform that directly uses the DLL handle. It's worth also noting that the DLL is delay-loaded on Windows so manually loading it shouldn't be necessary but I've left it in anyway.

Would love to see this merged, we've been running with these changes on Windows and Linux for over a month now :)

tustanivsky:
> Noticed an issue with the existing pathing to DLLs loaded at runtime - we didn't catch them since we don't test on Mac and that's seemingly the only platform that directly uses the DLL handle. It's worth also noting that the DLL is delay-loaded on Windows so manually loading it shouldn't be necessary but I've left it in anyway.
> 
> Would love to see this merged, we've been running with these changes on Windows and Linux for over a month now :)

Thank you for addressing that! I'll give it another round of testing later this week and let you know if any issues arise.

tustanivsky:
@VesCodes While testing these changes on Windows I've noticed that the plugin stopped logging events in the editor mode. What's more, it crashes the editor during the Sentry initialization in newly created projects (attaching the log). In packaged game builds everything works fine though.

Facing this issue on Windows 11 with UE 4.27.

I'll keep looking into this and let you know if anything else pops up.

[SentryPlayground.log](https://github.com/getsentry/sentry-unreal/files/11333855/SentryPlayground.log)

tustanivsky:
I've also tried to run this in our CI and even though plugin packaging works fine [tests are now failing].(https://github.com/getsentry/sentry-unreal/actions/runs/4817032377/jobs/8577261662#step:9:5738)

VesCodes:
Right, looks like the issue is that runtime dependencies don't get copied into "installed" directories. Since `$(TargetOutputDir)` is the engine binaries directory in a non-monolithic build (i.e. editor), when building against an installed distribution of the engine the dependencies don't get copied to that location.

@tustanivsky I'll have another whack at it and get back to you with a fix :)

VesCodes:
Hey @tustanivsky give this another go - tested this last change against a source engine and an installed engine in both editor and packaged Windows builds.

tustanivsky:
> Hey @tustanivsky give this another go - tested this last change against a source engine and an installed engine in both editor and packaged Windows builds.

Alright, the issues with Linux and Windows seem to be resolved however now I'm facing a similar crash on Mac. Currently it's a bit unclear whether this is a local problem of mine or something with the suggested code changes. I'll keep you posted

tustanivsky:
@VesCodes The change I've suggested is supposed to address the issue I've faced while testing the plugin on Mac. The thing is that the full path to `sentry.dylib` has to be specified in order to obtain a valid DLL handle.

I think it should give us a green light to merge the PR.

tustanivsky:
Superseded by #287

PR #273, opened 2023-05-11
Add plugin settings allowing to disable event capturing for certain build configurations/targets

### LINKED ISSUE #262: Ability to turn on & off platforms/configurations from sending crashes to Sentry
While we're currently waiting on #32 support, something that would be useful alongside this would to be able to turn off crashes for certain configurations or platforms, for example, we don't want to send Editor crashes to Sentry.

tustanivsky:
This change should provide extra flexibility for the Sentry initialization by allowing to disable it for certain build configurations/targets via the settings menu:

image

Closes #262

### DIFF
--- a/plugin-dev/Source/Sentry/Private/SentrySubsystem.cpp
+++ b/plugin-dev/Source/Sentry/Private/SentrySubsystem.cpp
@@ -4,6 +4,7 @@
 #include "SentryModule.h"
 #include "SentrySettings.h"
 #include "SentryBreadcrumb.h"
+#include "SentryDefines.h"
 #include "SentryEvent.h"
 #include "SentryId.h"
 #include "SentryUserFeedback.h"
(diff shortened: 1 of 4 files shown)


plugin-dev/Source/Sentry/Private/SentrySubsystem.cpp
```diff
@@ -338,3 +353,63 @@ void USentrySubsystem::DisableAutomaticBreadcrumbs()
 		FCoreDelegates::GameSessionIDChanged.Remove(GameSessionIDChangedDelegate);
 	}
 }
+
+bool USentrySubsystem::IsCurrentBuildConfigurationDisabled()
```
bitsandfoxes, line 357: Should this be moved behind the self-initialization?

bitsandfoxes (approved):
🔥

plugin-dev/Source/Sentry/Private/SentrySubsystem.cpp
```diff
@@ -47,13 +48,21 @@ void USentrySubsystem::Deinitialize()
 {
 	DisableAutomaticBreadcrumbs();
 
+	Close();
```
bitsandfoxes, line 51: Is that an earlier oversight?

tustanivsky: Yeah, noticed it while disabling the event capturing for Editor - without shutting down the native SDK upon subsystem's deinitialization events were still passing through in the subsequent game sessions even though the corresponding target was disabled

PR #294, opened 2023-06-06
Unify release name for all supported platforms

tustanivsky:
This fixes an issue with different release names on desktop/iOS/Android in case one wasn't provided explicitly in plugin settings.


bitsandfoxes (approved):
This looks great!

Issue #302, opened 2023-06-14
Crashes captured on Android/iOS have incorrect game log file attached


tustanivsky:
Whenever a crash occurs on Android/iOS and automatic game log attachment is enabled in plugin settings Sentry receives an incorrect file on the next app startup when the corresponding event is sent.

The game log which corresponds to the previous app session (one that crashed) gets renamed to `<ProjectName>-backup-<timestamp>.log` on application startup automatically while the native SDKs still looking for `<ProjectName>.log` that is now populated with the current session's info.

tustanivsky:
In order to resolve this issue for iOS https://github.com/getsentry/sentry-cocoa/issues/2325 has to be completed first.

PR #309, opened 2023-06-19
Fix automatic game log attachment (Android)

tustanivsky:
Related to #302

### DIFF
--- a/plugin-dev/Source/Sentry/Private/Android/Java/SentryBridgeJava.java
+++ b/plugin-dev/Source/Sentry/Private/Android/Java/SentryBridgeJava.java
@@ -3,14 +3,11 @@
 package io.sentry.unreal;
 
 import android.app.Activity;
-import android.content.pm.PackageInfo;
-import android.content.pm.PackageManager;
 
 import androidx.annotation.NonNull;
 
 import java.util.HashMap;
 import java.util.Map;
-import android.util.Log;
 
 import io.sentry.Attachment;
 import io.sentry.Breadcrumb;
--- a/plugin-dev/Source/Sentry/Private/Android/Java/SentryBridgeJava.java
+++ b/plugin-dev/Source/Sentry/Private/Android/Java/SentryBridgeJava.java
@@ -10,9 +10,11 @@ import androidx.annotation.NonNull;
 
 import java.util.HashMap;
 import java.util.Map;
+import android.util.Log;
 
 import io.sentry.Attachment;
 import io.sentry.Breadcrumb;
+import io.sentry.Hint;
 import io.sentry.IHub;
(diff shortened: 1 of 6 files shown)


plugin-dev/Source/Sentry/Private/Android/Java/SentryBridgeJava.java
```diff
@@ -45,15 +45,24 @@ public void configure(SentryAndroidOptions options) {
...
-			}
-		});
-
-		Sentry.configureScope(new ScopeCallback() {
-			@Override
-			public void run(@NonNull Scope scope) {
-				if(!gameLogPath.isEmpty()) {
-					scope.addAttachment(new Attachment(gameLogPath));
-				}
+				options.setBeforeSend(new SentryOptions.BeforeSendCallback() {
```
bitsandfoxes, line 48: I'm not entirely sure how that works underneath but can that callback be overwritten by users?

plugin-dev/Source/Sentry/Private/Android/Java/SentryBridgeJava.java
```diff
@@ -45,15 +45,24 @@ public void configure(SentryAndroidOptions options) {
...
+					@Override
+					public SentryEvent execute(SentryEvent event, Hint hint) {
+						if(event.isCrashed() && event.isErrored())
```
bitsandfoxes, line 51: Doesn't that mean that if the app crashed in the run before the gamelogs from the current run won't be part of any current events?

plugin-dev/Source/Sentry/Private/Android/Java/SentryBridgeJava.java
tustanivsky: We don't expose any APIs allowing users to do that, so it should be safe to set the `beforeSend` hook like this.

bitsandfoxes: Considering that we want to expose that API in the future, we could consider to add this as an eventprocessor/integration instead?

plugin-dev/Source/Sentry/Private/Android/Java/SentryBridgeJava.java
tustanivsky: No, after sending the crash event on app startup all subsequent events should be treated as usual with a current session's game log attached.

plugin-dev/Source/Sentry/Private/Android/Java/SentryBridgeJava.java
tustanivsky: Yeah, I think it should be feasible.

PR #327, opened 2023-07-16
Fixing issue where the Sentry.so file hasn't been copied yet to the binaries folder therefore resulting in a compile error

Edstub207:
Fixing the following:
`Library 'E:\Horde\Work\Blah\Sync\BLAH\Plugins\External\SentryUnrealPlugin\Binaries\Linux\libsentry.so' was not resolvable to a file when used in Module 'Sentry', assuming it is a filename and will search library paths for it. This is slow and dependency checking will not work for it. Please update reference to be fully qualified alternatively use PublicSystemLibraryPaths if you do intended to use this slow path to suppress this warning.  [...]`

While the RuntimeDependencies does mark this file correctly, it doesn't copy it in time for PublicAdditionalLibraries to use it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 - Add HTTP proxy for desktop ([#322](https://github.com/getsentry/sentry-unreal/pull/322))
 
+### Fixes
+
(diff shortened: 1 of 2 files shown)


Edstub207:
@tustanivsky FYI :)

PR #363, opened 2023-08-25
Add breakpad support for Windows

tustanivsky:
This PR switches `sentry-native` backend from `crashpad` to `breakpad`. It allows Sentry to comply with the [UE Marketplace publishing rules (2.6.2.e)](https://www.unrealengine.com/en-US/marketplace-guidelines#262e) which restrict the distribution of certain file types.

`breakpad` is an in-proc crash handler represented by a separate library unlike the `crashpad` which is essentially a separate app. The key difference it introduces for the crash-capturing flow is that game has to be re-launched in order to send captured crash to Sentry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 - Add extra crash context for native integration ([#342](https://github.com/getsentry/sentry-unreal/pull/342))
 - Add missing plugin settings ([#335](https://github.com/getsentry/sentry-unreal/pull/335))
 - Update event context categories for desktop ([#356](https://github.com/getsentry/sentry-unreal/pull/356))
-- Add breakpad support for Windows/Linux ([#363](https://github.com/getsentry/sentry-unreal/pull/363))
+- Add breakpad support for Windows ([#363](https://github.com/getsentry/sentry-unreal/pull/363))
 
 ### Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Add extra crash context for native integration ([#342](https://github.com/getsentry/sentry-unreal/pull/342))
 - Add missing plugin settings ([#335](https://github.com/getsentry/sentry-unreal/pull/335))
 - Update event context categories for desktop ([#356](https://github.com/getsentry/sentry-unreal/pull/356))
+- Add breakpad support for Windows/Linux ([#363](https://github.com/getsentry/sentry-unreal/pull/363))
 
 ### Fixes
 
--- a/plugin-dev/Source/Sentry/Sentry.Build.cs
+++ b/plugin-dev/Source/Sentry/Sentry.Build.cs
@@ -119,7 +119,6 @@ public class Sentry : ModuleRules
 			PublicRuntimeLibraryPaths.Add(PlatformBinariesPath);
 
 			PublicAdditionalLibraries.Add(Path.Combine(PlatformThirdPartyPath, "bin", "libsentry.so"));
-
 			PublicAdditionalLibraries.Add(Path.Combine(PlatformThirdPartyPath, "lib", "libbreakpad_client.a"));
 
 			PublicDefinitions.Add("USE_SENTRY_NATIVE=1");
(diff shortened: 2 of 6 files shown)


Edstub207:
@tustanivsky I'm intrigued RE: "The key difference it introduces for the crash-capturing flow is that game has to be re-launched in order to send captured crash to Sentry." - I can imagine there are cases where the process won't be restarted on the exact same code, or even potentially the local persistence might have been reset. For example, if a Linux Server is within a Docker Container and doesn't have a mounted volume for persistence or an update has been pushed since the title crashed.

Is this a hard limitation of Breakpad? Or is there something that can be done to ensure crashes are properly tracked in the above scenarios.

tustanivsky:
> Is this a hard limitation of Breakpad? Or is there something that can be done to ensure crashes are properly tracked in the above scenarios.

Yes, that's expected `breakpad` behavior which can't be changed by any means.

We haven't come up with the final decision regarding migrating to `breakpad` yet and this PR currently is more like a proof of concept thing.

Edstub207:
That makes sense, thanks for confirming @tustanivsky  - Once it's in a testable state, let me know, I'd be up for doing some testing comparing the two.

tustanivsky:
> That makes sense, thanks for confirming @tustanivsky - Once it's in a testable state, let me know, I'd be up for doing some testing comparing the two.

The branch is ready for testing so feel free to give `breakpad` a try.

Edstub207:
Thanks for confirming @tustanivsky - I've tried testing this locally and didn't see crashes uploaded. 

The data in the .sentry-native folder doesn't appear to demonstrate any crash reports and the last_crash file has 2023-08-17T10:26:24.830Z as the timestamp, which doesn't seem correct. 

However, on second boot of the client, after Sentry Init'd it did appear to try and send something, although, got a 404?

`[2023.08.29-23.08.58:294][  0]LogSentrySdk: received response:
HTTP/1.1 404 Not Found
Date: Tue, 29 Aug 2023 23:08:58 GMT
Via: 1.1 google
Content-Length: 14
Content-Type: text/plain
Server: nginx
Strict-Transport-Security: max-age=31536000; includeSubDomains; preload
Alt-Svc: h3=":443"; ma=2592000,h3-29=":443"; ma=2592000`

tustanivsky:
@tustanivsky Did you enable the `EnableAutoCrashCapturing` option in plugin settings? Also, that is supposed to work for UE 5.2+ only.

Edstub207:
> @tustanivsky Did you enable the `EnableAutoCrashCapturing` option in plugin settings? Also, that is supposed to work for UE 5.2+ only.

I'll do some testing around that today - Is that a requirement for breakpad?

tustanivsky:
> I'll do some testing around that today, I thought I did but maybe I missed it - Is that a requirement for breakpad?

Yes, the setup should be the same as it was for Crashpad.

Edstub207:
> > I'll do some testing around that today, I thought I did but maybe I missed it - Is that a requirement for breakpad?
> 
> Yes, the setup should be the same as it was for Crashpad.

Still been having some issues during testing, but I believe I spotted where the mistake was made earlier. Will keep on investigating.

Edstub207:
> > I'll do some testing around that today, I thought I did but maybe I missed it - Is that a requirement for breakpad?
> 
> Yes, the setup should be the same as it was for Crashpad.

@tustanivsky Based on further discussions with the team (And other projects as well) If breakpad is a route which is taken for this plugin, it might be worth having options to switch between Crashpad or Breakpad. There are some situations where breakpad is not suitable.

tustanivsky:
> Based on further discussions with the team (And other projects as well) If breakpad is a route which is taken for this plugin, it might be worth having options to switch between Crashpad or Breakpad. There are some situations where breakpad is not suitable.

@Edstub207 For now we've decided to move on with `breakpad` for Windows only and leave the Linux implementation without changes. Technically it should allow the plugin to finally make its way to the UE Marketplace while avoiding potential issues with Linux Server you've mentioned above. If that goes well we'll consider adding some mechanism giving users the ability to switch between crashpad/breakpad later.

PR #410, opened 2023-10-17
Fix crash capturing issue on Linux

tustanivsky:
This PR addresses the issue with crash reports not being uploaded to Sentry when `on_crash` hook is set during the `sentry-native` initialization on Linux. Whenever a crash occurs new instance of `FSharedCrashContext` struct is allocated on the signal's handler stack which might lead to its overflow and further corruption. Moving the crash context allocation to the heap instead resolves the problem and handler is able to process the captured event properly.

### DIFF
--- a/plugin-dev/Source/Sentry/Private/Desktop/CrashReporter/SentryCrashContext.cpp
+++ b/plugin-dev/Source/Sentry/Private/Desktop/CrashReporter/SentryCrashContext.cpp
@@ -9,7 +9,7 @@
 
 #if USE_SENTRY_NATIVE
 
-FSentryCrashContext::FSentryCrashContext(const FSharedCrashContext& Context)
+FSentryCrashContext::FSentryCrashContext(TSharedPtr<FSharedCrashContext> Context)
 #if ENGINE_MAJOR_VERSION == 5 && ENGINE_MINOR_VERSION >= 3
 	: CrashContext(Context)
 #else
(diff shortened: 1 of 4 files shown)


plugin-dev/Source/Sentry/Private/Desktop/CrashReporter/SentryCrashContext.cpp
```diff
@@ -31,15 +31,15 @@ void FSentryCrashContext::Apply(TSharedPtr<SentryScopeDesktop> Scope)
 {
 	const USentrySettings* Settings = FSentryModule::Get().GetSettings();
 
-	const FSessionContext& SessionContext = CrashContext.SessionContext;
+	const FSessionContext& SessionContext = CrashContext->SessionContext;
 
-	Scope->SetExtraValue("Crash Type", FGenericCrashContext::GetCrashTypeString(CrashContext.CrashType));
-	Scope->SetExtraValue("IsEnsure", CrashContext.CrashType == ECrashContextType::Ensure ? TEXT("true") : TEXT("false"));
+	Scope->SetExtraValue("Crash Type", FGenericCrashContext::GetCrashTypeString(CrashContext->CrashType));
```
bitsandfoxes, line 36: Worth noting that we should consider moving all the things that are getting set as `Extra` into their own context.

bitsandfoxes (approved):

plugin-dev/Source/Sentry/Private/Desktop/CrashReporter/SentryCrashContext.cpp
tustanivsky: Good point, let's track this in #411

PR #412, opened 2023-10-18
Fix event level on Win/Linux

### LINKED ISSUE #409: Events always have `Debug` level on Windows/Linux
On Win/Linux plugin uses a custom `Scope` implementation which overrides the event's level within the `BeforeSend` handler:

```
sentry_value_t HandleBeforeSend(sentry_value_t event, void *hint, void *closure)
{
    SentrySubsystemDesktop* SentrySubsystem = static_cast<SentrySubsystemDesktop*>(closure);

    USentryEvent* EventToProcess = NewObject<USentryEvent>();
    EventToProcess->InitWithNativeImpl(MakeShareable(new SentryEventDesktop(event)));

    SentrySubsystem->GetCurrentScope()->Apply(EventToProcess);

    return SentrySubsystem->GetBeforeSendHandler()->HandleBeforeSend(EventToProcess, nullptr) ? event : sentry_value_new_null();
}
```
This leads to a situation when all captured events have the `Debug` level (set by default) or any other level set for the global scope ignoring one that was specified upon event creation.

tustanivsky:
This PR addresses the issue with events' level always being overwritten by a current scope's level (`Debug` by default) on Win/Linux. Now scope's level is applied only to the events that don't have this property set, otherwise events' level remains unchanged.

Also, the convenience function for creating event objects was mo moved to the `USentryEvent` class + function allowing to create an empty event (without a pre-set level) was added.

Closes #409

### DIFF
--- a/plugin-dev/Source/Sentry/Private/SentryEvent.cpp
+++ b/plugin-dev/Source/Sentry/Private/SentryEvent.cpp
@@ -26,11 +26,6 @@ USentryEvent::USentryEvent()
 	}
 }
 
-USentryEvent* USentryEvent::CreateEmptyEvent()
-{
-	return NewObject<USentryEvent>();
-}
-
 USentryEvent* USentryEvent::CreateEventWithMessageAndLevel(const FString& Message, ESentryLevel Level)
 {
 	USentryEvent* Event = NewObject<USentryEvent>();
--- a/plugin-dev/Source/Sentry/Private/SentryEvent.cpp
+++ b/plugin-dev/Source/Sentry/Private/SentryEvent.cpp
@@ -36,7 +36,9 @@ USentryEvent* USentryEvent::CreateEventWithMessageAndLevel(const FString& Messag
 	USentryEvent* Event = NewObject<USentryEvent>();
 
 	if(!Message.IsEmpty())
+	{
 		Event->SetMessage(Message);
+	}
 
 	Event->SetLevel(Level);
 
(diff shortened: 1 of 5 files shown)


plugin-dev/Source/Sentry/Public/SentryEvent.h
```diff
@@ -19,6 +19,19 @@ class SENTRY_API USentryEvent : public UObject
 public:
 	USentryEvent();
 
+	/** Creates the empty event. */
+	UFUNCTION(BlueprintCallable, Category = "Sentry")
+	static USentryEvent* CreateEmptyEvent();
```
bitsandfoxes, line 24:
```suggestion
	static USentryEvent* CreateEvent();
```
It's unfortunate that we can't overload functions for blueprints.

plugin-dev/Source/Sentry/Private/SentryEvent.cpp
```diff
@@ -26,6 +26,23 @@ USentryEvent::USentryEvent()
...
+{
+	return NewObject<USentryEvent>();
+}
+
+USentryEvent* USentryEvent::CreateEventWithMessageAndLevel(const FString& Message, ESentryLevel Level)
+{
+	USentryEvent* Event = NewObject<USentryEvent>();
+
+	if(!Message.IsEmpty())
+		Event->SetMessage(Message);
```
bitsandfoxes, line 39:
```suggestion
	if(!Message.IsEmpty())
	{
		Event->SetMessage(Message);
	}
```
Just to stay consistent with the rest.

bitsandfoxes (approved):

plugin-dev/Source/Sentry/Public/SentryEvent.h
tustanivsky: After giving it a second thought I guess we might actually remove that completely and simply create a new event using `NewObject` (in CPP) and `ConstructObject` (in BP)

PR #429, opened 2023-10-26
Fix iOS build errors

tustanivsky:
This addresses build errors that arised presumably after merging #417  (UIKit linking)


bitsandfoxes (approved):
Thanks! Nice catch!

PR #433, opened 2023-11-03
Switch to using static libs on Windows

tustanivsky:
Switching to the static version of `sentry-native` libs on Windows should allow bypassing issues with publishing to UE Marketplace (obfuscation of certain binaries required by Sentry plugin).

Related to #424


.github/workflows/ci.yml
```diff
@@ -33,13 +33,13 @@ jobs:
     uses: ./.github/workflows/sdk.yml
     with:
       target: Win64-Crashpad
-      runsOn: windows-latest
+      runsOn: windows-2019
```
bitsandfoxes, line 36: Why is that necessary?

bitsandfoxes (commented):
Why do we need specifically `VS 2019` to build sentry-native?

tustanivsky:
> Why do we need specifically `VS 2019` to build sentry-native?

In order to preserve the [compatibility](https://docs.unrealengine.com/5.3/en-US/setting-up-visual-studio-development-environment-for-cplusplus-projects-in-unreal-engine/) of `sentry-native` static libs with Unreal we have to stick to VS 2019. Afaik VS 2019 version remains available only at `windows-2019` VMs so that's why we have to hold up with using `windows-latest` at this point.

bitsandfoxes (approved):

PR #443, opened 2023-11-15
Fix readme

GloryOfNight:
crashpad_handler has additional libcurl runtime dependency

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ The SDK complies with three latest engine versions.
 
 - Only crash events captured on Android contain the full callstack. Events that were captured manually won't have the native C++ part there.
 
-- On Linux `sudo apt-get install libc++-dev libcurl4` is required to install the required `Crashpad` dependencies.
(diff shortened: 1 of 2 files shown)


GloryOfNight:
For us crashpad_handler did not required libc++-dev to be installed, maybe it's no longer needed. Someone should check for that

tustanivsky:
> For us crashpad_handler did not require libc++-dev to be installed, maybe it's no longer needed. Someone should check for that

Can you elaborate on what's your current Linux setup (distro, version, etc.)? I don't recall encountering an issue with the missing `libcurl` before so maybe those extra bits of info will help us understand whether it's a common problem or something specific to your environment

GloryOfNight:
We running our servers inside a container, using:
https://github.com/orgs/epicgames/packages/container/unreal-engine/111082761?tag=runtime

And distro according to server logs would be: Ubuntu 18.04.6 LTS (5.15.120+)

tustanivsky:
Superseded by #455

PR #464, opened 2023-12-11
Fix Android build errors after upgrading to `sentry-java` version `0.7.0`

tustanivsky:
This PR replaces `Scope` class with a corresponding `IScope` interface in plugin's Java implementation to fix build errors.

PR #465, opened 2023-12-11
Fix Android build error after integrated to 0.15.0

TroutZhang:
Related to change https://github.com/getsentry/sentry-java/pull/3066


bitsandfoxes:
Hey @TroutZhang, thanks for the PR. Unlucky timing but we already merged https://github.com/getsentry/sentry-unreal/pull/464
The fix is on its way.

PR #468, opened 2023-12-18
Fix Linux intermediates paths in `FilterPlugin.ini`

tustanivsky:
Related to #453

### DIFF
--- a/plugin-dev/Config/FilterPlugin.ini
+++ b/plugin-dev/Config/FilterPlugin.ini
@@ -7,5 +7,10 @@
 ;    /Extras/...
 ;    /Binaries/ThirdParty/*.dll
 /Scripts/*
-/LinuxIntermediate/*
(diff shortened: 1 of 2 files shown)


plugin-dev/Config/FilterPlugin.ini
```diff
@@ -7,5 +7,10 @@
 ;    /Extras/...
 ;    /Binaries/ThirdParty/*.dll
 /Scripts/*
-/LinuxIntermediate/*
```
bitsandfoxes, line 10: Why doesn't this work?

tustanivsky: Apparently the `*` wildcard doesn't prevent UE Marketplace from obfuscating subdirectories in `LinuxIntermediate` folder during the publishing so I assume that specifying those explicitly should help

bitsandfoxes (approved):

PR #479, opened 2024-01-10
Add flag indicating which package version is used (GitHub or Marketplace)

tustanivsky:
This PR introduces the following changes:
- `IsMarketplace` flag value is set in CI during the plugin packaging
- `IsMarketplace` flag added to the default event's context
- Updated `sentry-native` initialization logic
- Fixed issue with permanently disabling crash reporter


plugin-dev/Source/Sentry/Private/Desktop/SentrySubsystemDesktop.cpp
```diff
@@ -104,11 +95,24 @@ void SentrySubsystemDesktop::InitWithSettings(const USentrySettings* settings, U
 	}
 
 #if PLATFORM_WINDOWS
-	sentry_options_set_handler_pathw(options, *FPaths::ConvertRelativePathToFull(HandlerPath));
-	sentry_options_set_database_pathw(options, *FPaths::ConvertRelativePathToFull(DatabasePath));
+	if(!FSentryModule::Get().IsMarketplaceVersion())
+	{
+		if(!FPaths::FileExists(GetHandlerPath()))
+		{
+			UE_LOG(LogSentrySdk, Log, TEXT("Crashpad executable couldn't be found so Breakpad will be used instead. "
```
bitsandfoxes, line 102: I don't get how we fall back to `Breakpad`. Is that an implicit fallback during initialization due to a missing `Crashpad` executable? Because the non-existing path still gets set.

plugin-dev/Source/Sentry/Private/Desktop/SentrySubsystemDesktop.cpp
```diff
@@ -136,10 +140,9 @@ void SentrySubsystemDesktop::InitWithSettings(const USentrySettings* settings, U
 	isEnabled = initResult == 0 ? true : false;
 
 #if PLATFORM_WINDOWS && ENGINE_MAJOR_VERSION == 5 && ENGINE_MINOR_VERSION >= 2
-	if(settings->EnableAutoCrashCapturing)
-	{
-		FPlatformMisc::SetCrashHandlingType(ECrashHandlingType::Disabled);
-	}
+	FPlatformMisc::SetCrashHandlingType(settings->EnableAutoCrashCapturing
+		? ECrashHandlingType::Disabled
+		: ECrashHandlingType::Default);
```
bitsandfoxes, line 145: This is a fallback to still keep the built-in crash reporter functional?

plugin-dev/Source/Sentry/Private/Desktop/SentrySubsystemDesktop.cpp
```diff
@@ -104,11 +95,24 @@ void SentrySubsystemDesktop::InitWithSettings(const USentrySettings* settings, U
 	}
 
 #if PLATFORM_WINDOWS
-	sentry_options_set_handler_pathw(options, *FPaths::ConvertRelativePathToFull(HandlerPath));
-	sentry_options_set_database_pathw(options, *FPaths::ConvertRelativePathToFull(DatabasePath));
+	if(!FSentryModule::Get().IsMarketplaceVersion())
+	{
+		if(!FPaths::FileExists(GetHandlerPath()))
```
bitsandfoxes, line 100: This is probably a nitpick but can we keep the path as a local variable so we don't need to `FPaths::Combine` and `ToFullPahth` twice? Here and where we actually set it?

plugin-dev/Source/Sentry/Private/Desktop/SentrySubsystemDesktop.cpp
tustanivsky: Yes, without it we won't be able to restore normal crash reporter behavior within the same editor session even after disabling `EnableAutoCrashCapturing` in plugin's settings.

plugin-dev/Source/Sentry/Private/Desktop/SentrySubsystemDesktop.cpp
tustanivsky: I've put back the local variable within the `if(!FSentryModule::Get().IsMarketplaceVersion())` block here to avoid calling those path utils twice.

bitsandfoxes (approved):

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Features
 
+- Add flag indicating which package version is used (GitHub or Marketplace) ([#479](https://github.com/getsentry/sentry-unreal/pull/479))
```
bitsandfoxes, line 7:
```suggestion
- Improved crash capture backend handling based on package version (GitHub or Marketplace) ([#479](https://github.com/getsentry/sentry-unreal/pull/479))
```

PR #488, opened 2024-01-29
Add traces sampler function config for Android/Apple

tustanivsky:
With this PR it becomes possible to specify traces sampler function in plugin settings which allows determining traces sample rate based on the sampling context (alternative to fixed sample rate).


plugin-dev/Source/Sentry/Private/Android/Java/SentryBridgeJava.java
```diff
@@ -73,7 +73,12 @@ public SentryEvent execute(SentryEvent event, Hint hint) {
 						options.setTracesSampler(new SentryOptions.TracesSamplerCallback() {
 							@Override
 							public Double sample(SamplingContext samplingContext) {
-								return onTracesSampler(samplerAddr, samplingContext);
+								float sampleRate = onTracesSampler(samplerAddr, samplingContext);
+								if(sampleRate > 0.0f) {
```
bitsandfoxes, line 77:
```suggestion
								if(sampleRate >= 0.0f) {
```

bitsandfoxes (approved):

plugin-dev/Source/Sentry/Private/Android/Java/SentryBridgeJava.java
tustanivsky: Good catch!

PR #499, opened 2024-02-21
Fix casing in include

KieranChandler:
For some reason this particular file has inconsistent casing with the rest of the files in the UE4 codebase and Linux is obviously case sensitive, so this causes build failures.

I've updated the include to match the casing of the actual file from UE4

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 ### Fixes
 
 - Fix Linux intermediates paths in `FilterPlugin.ini` ([#468](https://github.com/getsentry/sentry-unreal/pull/468))
+- Fix casing for include of HAL/PlatformFileManager for Linux compilation ([#468](https://github.com/getsentry/sentry-unreal/pull/499))
 
 ### Dependencies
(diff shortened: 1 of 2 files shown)


tustanivsky:
@KieranChandler Will this still work in UE5 though?

I guess adding some extra checks might be worth consideration:
```cpp
#if ENGINE_MAJOR_VERSION >= 5
#include "HAL/PlatformFileManager.h"
#else
#include "HAL/PlatformFilemanager.h"
#endif
```

KieranChandler:
I can check the casing of the file in UE5 and if its different then i could do that

KieranChandler:
Its good you asked that. because they fixed it in 5.0: https://github.com/EpicGames/UnrealEngine/commit/57baf1e5c77a7e6d26537927a705316ef67a4452

tustanivsky (approved):

tustanivsky:
@KieranChandler Please add a corresponding changelog entry and I think we should be good to merge this. Thank you for bringing it up!

Issue #511, opened 2024-03-05
Add a cpp only version of CaptureMessageWithScope


Maxmystere:
I would like to have a version of CaptureMessageWithScope that would take a `DECLARE_DELEGATE_OneParam` instead of the current `DECLARE_DYNAMIC_DELEGATE_OneParam` so that it could be used directly in cpp with lambda instead of having to rely on an UObject to store my data

If the Feature Request is validated I can try to do it

tustanivsky:
Thank you, we'll review this feature request and get back to you shortly.

tustanivsky:
@Maxmystere The above request looks reasonable and we fully support the idea of making the plugin's API more convenient for those who use it in C++. I can't provide any ETA when we'll be up to implementing this on our side however if you manage to come up with a solution please feel free to open a corresponding PR.

bitsandfoxes:
Closed by https://github.com/getsentry/sentry-unreal/pull/516

PR #516, opened 2024-03-13
Implement not dynamic delegate for ConfigureScope #511

Maxmystere:
Here is my PR for implementing the use of DECLARE_DELEGATE_OneParam for ConfigureScope #511

Converting FConfigureScopeDelegate to a non dynamic delegate might cause issues on migration of code for existing cpp users as now the cpp version of the implementation would be `FConfigureScopeDynDelegate` (but maybe could be good to make aware for people to replace their existing cpp code to use `FConfigureScopeDelegate`

### DIFF
--- a/plugin-dev/Source/Sentry/Private/Android/Callbacks/SentryScopeCallbackAndroid.cpp
+++ b/plugin-dev/Source/Sentry/Private/Android/Callbacks/SentryScopeCallbackAndroid.cpp
@@ -10,7 +10,7 @@ USentryScopeCallbackAndroid::USentryScopeCallbackAndroid()
 	}
 }
 
-void USentryScopeCallbackAndroid::BindDelegate(const FConfigureScopeDelegate& OnConfigure)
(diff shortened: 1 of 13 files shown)


bitsandfoxes:
Thanks for contributing and sorry for the delay, we'll get back to you as soon as possible.

Maxmystere:
Thanks for the reply, I got busy with other tasks so I didn't get the time to test it on mobile platforms but it should be ready for tests

tustanivsky:
@Maxmystere I've run some tests for this on my side and everything works perfectly. The only thing I'd like to suggest is to rename the newly added non-dynamic delegate to `FConfigureScopeNativeDelegate` instead of changing the existing dynamic one - it should allow us to avoid breaking things for those who are using the plugin's API in C++.

Maxmystere:
Thanks for the tests ! I've updated with what you asked, makes the review a bit bigger but yeah it's cleaner for people updating

tustanivsky:
> Thanks for the tests ! I've updated with what you asked, makes the review a bit bigger but yeah it's cleaner for people updating

Great! I'll give it another go and get back to you.

tustanivsky (approved):

PR #522, opened 2024-03-25
Add breadcrumbs automatically when printing to logs

### LINKED ISSUE #382: Support logs as breadcrumbs
We are in the process of migrating from our own unreal sentry plugin, which we have been using for a couple of years, 
https://github.com/mainframeindustries/UE4SentryClientPlugin

One nice feature it has it that it hooks into the unreal log handler to gather breadcrumbs that are added to sentry events:

https://github.com/mainframeindustries/UE4SentryClientPlugin/blob/3a3b8600c99135a17c18ab2a2fbd44df32ba6590/Source/SentryClient/Private/SentryClientModule.cpp#L450C1-L463

Is this something which could be easily added?

tustanivsky:
Closes #382

### DIFF
--- a/plugin-dev/Source/Sentry/Private/SentryOutputDevice.cpp
+++ b/plugin-dev/Source/Sentry/Private/SentryOutputDevice.cpp
@@ -1,0 +1,73 @@
+﻿// Copyright (c) 2024 Sentry. All Rights Reserved.
+
+#include "SentryOutputDevice.h"
+
+#include "SentryBreadcrumb.h"
+#include "SentryModule.h"
+#include "SentrySettings.h"
+#include "SentrySubsystem.h"
+
+void FSentryOutputDevice::Serialize(const TCHAR* V, ELogVerbosity::Type Verbosity, const FName& Category)
+{
+	const USentrySettings* Settings = FSentryModule::Get().GetSettings();
+
+	bool bAddBreadcrumb;
+
+	ESentryLevel BreadcrumbLevel = ESentryLevel::Debug;
+
+	switch (Verbosity)
+	{
+	case ELogVerbosity::Fatal:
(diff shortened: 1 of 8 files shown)


plugin-dev/Source/Sentry/Private/SentryOutputDevice.cpp
```diff
@@ -0,0 +1,75 @@
...
+		bAddBreadcrumb = false;
+	}
+
+	if(!bAddBreadcrumb)
+	{
+		return;
+	}
+
+	USentrySubsystem* SentrySubsystem = GEngine->GetEngineSubsystem<USentrySubsystem>();
+	if(!SentrySubsystem || !SentrySubsystem->IsEnabled())
```
bitsandfoxes, line 54: Just to confirm, it was because of the spaces!

plugin-dev/Source/Sentry/Private/SentrySubsystem.cpp
```diff
@@ -115,6 +117,13 @@ void USentrySubsystem::Initialize()
 
 	PromoteTags();
 	ConfigureBreadcrumbs();
+
+	OutputDevice = MakeShareable(new FSentryOutputDevice());
+	if(OutputDevice)
+	{
+		GLog->AddOutputDevice(OutputDevice.Get());
+		GLog->SerializeBacklog(OutputDevice.Get());
```
bitsandfoxes, line 125: Is there a chance for this to get overwritten by the user or something else?

bitsandfoxes (approved):
There's even the configuration of what types of breadcrumbs as well! Nice!

plugin-dev/Source/Sentry/Private/SentrySubsystem.cpp
tustanivsky: I don't think so - having multiple output devices is totally acceptable as they don't interfere with each other.

PR #536, opened 2024-04-10
Map sentry_level_t to ELogVerbosity::Type in PrintVerboseLog.

donalmacc:
Sentry's logging writes verbose SDK information to the default log level. I'd like to suggest mapping the sentry_level_t parameter to ELogVerbosity::Type.

Three potential things to raise here:
-  The fallback behaviour logs to the Error Category, alongside a message for an unknown sentry_level. 
- SENTRY_LEVEL_FATAL logs will shut down the app. In my testing, I've never seen a SENTRY_LEVEL_FATAL log message so I don't know what the right thing to do here is. I've mapped 1:1 to Unreal's categories, but I think logging to Error with a wrapper message might be a good idea.
- It might be worth having a Static Log Category for SDK messages specifically, something like:

`DEFINE_LOG_CATEGORY_STATIC(LogSentryNativeSdk, Display, All);` and then using that _just_ for these messages.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+- Map sentry_level_t to ELogVerbosity::Type in PrintVerboseLog ([#536](https://github.com/getsentry/sentry-unreal/pull/536))
+
+
 ### Features
 
 - Add user feedback capturing support for desktop ([#521](https://github.com/getsentry/sentry-unreal/pull/521))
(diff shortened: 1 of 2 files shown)


tustanivsky:
Looks great, thank you for this contribution!

Just a few comments regarding the above points:
- 1:1 mapping should work just fine. Recently we've introduced functionality that does the opposite conversion so extracting both things to separate utility functions will make sense as well.
- For now we'd prefer not to introduce additional log categories to keep things consistent across different platforms.

tustanivsky (approved):

PR #537, opened 2024-04-10
Fix invalid assertion issues grouping

### LINKED ISSUE #1: UE4 Asserts not captured on Android

### LINKED ISSUE #489: Several Stacktraces can ignored when an event was reported
We made our own delegate to report the Android platform's assertions，But the top couple of Stacktraces we don't want to report，or that's what happens：we want the "UEngine:PerformError" to be top so  that the title is displayed correctly. 
Otherwise all assertions are displayed “SentrySubsystemAndroid::OnCrashing” ，[Issue Grouping]--[FINGERPRINT RULES]--[stack.*] only get the top stack so i cannot make a correctly title
image
Or is there another way to solve this problem？
thanks~

### LINKED ISSUE #514: When an assert fires FGenericPlatformMisc::RaiseException is tracked instead than the actual error
### Environment

How do you use Sentry?
Sentry SaaS (sentry.io)

Which version of the SDK?
0.16.0

Git Download

Which version of Unreal?
5.3.2

Linux Server

### Steps to Reproduce

1. Run a server
2. Observe that instead than the assert FGenericPlatformMisc::RaiseException is tracked as the issue instead

### Expected Result
The assert is the captured event in Sentry 

### Actual Result
The default raise exception is captured and the assert is "lost" 

### Any logs or screenshots

Log cuts off at: 
`[2024.03.12-14.00.37:849][466]LogOutputDevice: Warning: `

When just after that the following occurs on the actual session log:
```
[2024.03.12-15.48.57:139][448]LogSentrySdk: flushing session and queue before crashpad handler
[2024.03.12-15.48.57:139][448]LogSentrySdk: invoking `on_crash` hook
[2024.03.12-15.48.57:140][448]LogSentrySdk: sending envelope
[2024.03.12-15.48.57:140][448]LogSentrySdk: serializing envelope into buffer
[2024.03.12-15.48.57:140][448]LogSentrySdk: handing control over to crashpad
CommonUnixCrashHandler: Signal=11
[2024.03.12-15.48.57:300][448]LogCore: === Critical error: ===
Unhandled Exception: SIGSEGV: invalid attempt to read memory at address 0x0000000000000008

[2024.03.12-15.48.57:300][448]LogCore: Fatal error!
```

tustanivsky:
This PR adds a custom `FOutputDeviceError` implementation which should allow to hook into Unreal's error handling flow and override assert reporting logic.

By default, an exception is raised whenever assert is fired, which is then captured by `crashpad`/`breakpad` backend. The topmost callstack frames for such events are always the same preventing them from being grouped properly and they all appear under a single issue at Sentry.

To avoid this situation instead of crashing the app a corresponding event can be constructed manually and reported via `CaptureEvent`. Once it's done the app is forced to shut down which somewhat repeats the original flow.

The current problem with the approach suggested here is that the engine's error handling is mixed up with some platform-specific logic (i.e. logging, restoring UI, etc.) which leads to skipping some parts of it when using the new `FSentryOutputDeviceError`.

Closes #489
Closes #514
Closes #1


tustanivsky:
On Android assertions are intercepted by `FSentryOutputDeviceError` which also sets several global tags used for further analysis during the `beforeSend` on the next application launch. Those tags allow us to identify whether the assertion occurred and how many call stack frames must be cut off. This change also fixes the #1

Edstub207:
@tustanivsky is there an update on this change?

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 
 ### Fixes
 
+- Fix invalid assertion issues grouping ([#537](https://github.com/getsentry/sentry-unreal/pull/537))
```
bitsandfoxes, line 12:
```suggestion
- The SDK now correctly captures and groups Assertions ([#537](https://github.com/getsentry/sentry-unreal/pull/537))
```

bitsandfoxes: But this almost feels like a feature to me.

plugin-dev/Source/Sentry/Private/Android/Java/SentryBridgeJava.java
```diff
@@ -89,6 +94,24 @@ public Double sample(SamplingContext samplingContext) {
 		});
 	}
 
+	private static void checkForUnrealException(SentryEvent event) {
```
bitsandfoxes, line 97: This does a whole lot more than just check for an Unreal exception as it returns nothing but modifies an event.

plugin-dev/Source/Sentry/Private/Android/Java/SentryBridgeJava.java
```diff
@@ -53,7 +57,8 @@ public void configure(SentryAndroidOptions options) {
 					options.setBeforeSend(new SentryOptions.BeforeSendCallback() {
 						@Override
 						public SentryEvent execute(SentryEvent event, Hint hint) {
-							return onBeforeSend(beforeSendHandler, event, hint);
+						checkForUnrealException(event);
```
bitsandfoxes, line 60: This is fairly messy. We modify an event based on tags the SDK should not set in the first place. And all of that happens in the java bridge code. Can we move this into a processor instead?

plugin-dev/Source/Sentry/Private/Android/Java/SentryBridgeJava.java
```diff
@@ -89,6 +94,24 @@ public Double sample(SamplingContext samplingContext) {
...
+				exception.setValue(event.getTag("sentry_unreal_exception_message"));
+				SentryStackTrace trace = exception.getStacktrace();
+				int numFramesToSkip = Integer.parseInt(event.getTag("sentry_unreal_exception_skip_frames"));
+				List<SentryStackFrame> frames = trace.getFrames();
+				trace.setFrames(frames.subList(0, frames.size() - numFramesToSkip));
+			}
+			event.removeTag("sentry_unreal_exception_type");
+			event.removeTag("sentry_unreal_exception_message");
+			event.removeTag("sentry_unreal_exception_skip_frames");
+			event.removeTag("sentry_unreal_exception");
```
bitsandfoxes, line 111: We really have to reevaluate the use of tags in the SDK. As a general rule, the SDK should not set any tags and add that information in contexts instead. Relay then handles those and elevates specific ones to tags.

plugin-dev/Source/Sentry/Private/Android/SentrySubsystemAndroid.cpp
```diff
@@ -169,6 +169,32 @@ USentryId* SentrySubsystemAndroid::CaptureEventWithScope(USentryEvent* event, co
 	return SentryConvertorsAndroid::SentryIdToUnreal(*id);
 }
 
+USentryId* SentrySubsystemAndroid::CaptureException(const FString& type, const FString& message, int32 framesToSkip)
+{
+	// add marker tags specific for Unreal assertions
+	SetTag(TEXT("sentry_unreal_exception"), TEXT("assert"));
+	SetTag(TEXT("sentry_unreal_exception_skip_frames"), FString::Printf(TEXT("%d"), framesToSkip));
+	SetTag(TEXT("sentry_unreal_exception_type"), type);
+	SetTag(TEXT("sentry_unreal_exception_message"), message);
```
bitsandfoxes, line 178: Can these be put in `contexts` instead?

plugin-dev/Source/Sentry/Private/Android/SentrySubsystemAndroid.cpp
```diff
@@ -169,6 +169,32 @@ USentryId* SentrySubsystemAndroid::CaptureEventWithScope(USentryEvent* event, co
...
+}
+
+USentryId* SentrySubsystemAndroid::CaptureAssertion(const FString& type, const FString& message)
+{
+	return CaptureException(type, message, 8);
+}
+
+USentryId* SentrySubsystemAndroid::CaptureEnsure(const FString& type, const FString& message)
+{
+	auto id = FSentryJavaObjectWrapper::CallStaticObjectMethod<jobject>(SentryJavaClasses::SentryBridgeJava, "captureException", "(Ljava/lang/String;Ljava/lang/String;)Lio/sentry/protocol/SentryId;",
```
bitsandfoxes, line 192: I don't follow. Why can we call `CaptureException` from in `CaptureAssertion` but we have to go straight to the bridge?

plugin-dev/Source/Sentry/Private/Android/SentrySubsystemAndroid.cpp
```diff
@@ -169,6 +169,32 @@ USentryId* SentrySubsystemAndroid::CaptureEventWithScope(USentryEvent* event, co
 	return SentryConvertorsAndroid::SentryIdToUnreal(*id);
 }
 
+USentryId* SentrySubsystemAndroid::CaptureException(const FString& type, const FString& message, int32 framesToSkip)
```
bitsandfoxes, line 172: I don't follow this. `CaptureException` does not actually capture an exception. Instead it adds a whole bunch of markers that are later getting picked up by the bridge?

plugin-dev/Source/Sentry/Private/Android/SentrySubsystemAndroid.cpp
```diff
@@ -169,6 +169,32 @@ USentryId* SentrySubsystemAndroid::CaptureEventWithScope(USentryEvent* event, co
...
+	SetTag(TEXT("sentry_unreal_exception_message"), message);
+
+	PLATFORM_BREAK();
+
+	return nullptr;
+}
+
+USentryId* SentrySubsystemAndroid::CaptureAssertion(const FString& type, const FString& message)
+{
+	return CaptureException(type, message, 8);
```
bitsandfoxes, line 187: Alright, so assertions and exceptions are getting treated similarly. Why the hardcoded `8`? Is that the same for all UE versions on all Android versions?

plugin-dev/Source/Sentry/Private/Android/SentrySubsystemAndroid.h
```diff
@@ -18,6 +18,9 @@ class SentrySubsystemAndroid : public ISentrySubsystem
 	virtual USentryId* CaptureMessageWithScope(const FString& message, const FConfigureScopeNativeDelegate& onConfigureScope, ESentryLevel level) override;
 	virtual USentryId* CaptureEvent(USentryEvent* event) override;
 	virtual USentryId* CaptureEventWithScope(USentryEvent* event, const FConfigureScopeNativeDelegate& onConfigureScope) override;
+	virtual USentryId* CaptureException(const FString& type, const FString& message, int32 framesToSkip) override;
```
bitsandfoxes, line 21: That's a public API, right? Should we really require `framesToSkip` from users?

plugin-dev/Source/Sentry/Private/Desktop/SentrySubsystemDesktop.cpp
```diff
@@ -153,6 +153,7 @@ void SentrySubsystemDesktop::InitWithSettings(const USentrySettings* settings, U
 	sentry_options_set_max_breadcrumbs(options, settings->MaxBreadcrumbs);
 	sentry_options_set_before_send(options, HandleBeforeSend, this);
 	sentry_options_set_on_crash(options, HandleBeforeCrash, this);
+	sentry_options_set_shutdown_timeout(options, 3000);
```
bitsandfoxes, line 156: Where is this coming from?

plugin-dev/Source/Sentry/Private/SentrySubsystem.cpp
```diff
@@ -610,3 +615,32 @@ bool USentrySubsystem::IsPromotedBuildsOnlyEnabled()
...
+}
+
+void USentrySubsystem::ConfigureOutputDeviceError()
+{
+	OutputDeviceError = MakeShareable(new FSentryOutputDeviceError(GError));
+	if (OutputDeviceError)
+	{
+		OutputDeviceError->OnError.AddLambda([this](const FString& Message)
+		{
+			SubsystemNativeImpl->CaptureAssertion(TEXT("Assertion failed"), Message);
```
bitsandfoxes, line 636: How do we know here, based on the checks, that this is an assertion and not a different kind of error?

bitsandfoxes (changes requested):
Thanks for all the effort put into this! It'll be a huge QOL improvement!
There are just a couple of things that concern me:
1. Addition to `CaptureException/Assertion/Ensure` to the public API. As far as I can tell they are getting used by the `OutputDevice`? But since they are public and user facing they don't make much sense to me - especially with the `skip frames` in the signature.
2. I'm not a huge fan of modifying the stacktrace on the client to "clean it up" as this gets messy easily. Updates and changes to the engine might require changes to the SDK which will require users to update their project e.t.c. So ideally, this would be done by the server. Since this works and we don't have an alternative in place I'm fine with adding this to the SDK but it should be done in a way that's easy for us to remove/replace. And it should be done away from anything `public`.

plugin-dev/Source/Sentry/Private/SentrySubsystem.cpp
tustanivsky: The `FSentryOutputDeviceError` makes the `FDebug::HasAsserted()` check when an error occurs and fires the `OnError` event only if it's true.

plugin-dev/Source/Sentry/Private/Android/SentrySubsystemAndroid.h
tustanivsky: No, we have `CaptureException` for internal use only to report asserts/ensures.

plugin-dev/Source/Sentry/Private/Desktop/SentrySubsystemDesktop.cpp
tustanivsky: Previously, UE caused the app to crash whenever an assertion failed and the Crashpad captured these crashes for us. We've changed how assertions are handled so now when it fails a timeout is used to ensure a corresponding event is sent to Sentry within the same session before shutting down the app manually.

plugin-dev/Source/Sentry/Private/Android/Java/SentryBridgeJava.java
```diff
@@ -117,12 +140,22 @@ public static SentryId captureEventWithScope(final SentryEvent event, final long
 		SentryId eventId = Sentry.captureEvent(event, new ScopeCallback() {
 			@Override
 			public void run(@NonNull IScope scope) {
-				onConfigureScope(callback, scope);
+			onConfigureScope(callback, scope);
```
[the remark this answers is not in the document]
tustanivsky: No, we don't. I guess we should consider adding a linter as the next improvement for our CI.

plugin-dev/Source/Sentry/Private/Android/Java/SentryBridgeJava.java
```diff
@@ -73,12 +78,12 @@ public SentryEvent execute(SentryEvent event, Hint hint) {
 						options.setTracesSampler(new SentryOptions.TracesSamplerCallback() {
 							@Override
 							public Double sample(SamplingContext samplingContext) {
-								float sampleRate = onTracesSampler(samplerAddr, samplingContext);
-								if(sampleRate >= 0.0f) {
-									return (double) sampleRate;
-								} else {
-									return null;
-								}
```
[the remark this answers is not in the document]
tustanivsky: Yes, there were quite a few formatting issues in that Java bridge script.

plugin-dev/Source/Sentry/Private/Android/Java/SentryBridgeJava.java
```diff
@@ -106,8 +129,8 @@ public static SentryId captureMessageWithScope(final String message, final Sentr
 		SentryId messageId = Sentry.captureMessage(message, new ScopeCallback() {
 			@Override
 			public void run(@NonNull IScope scope) {
-				scope.setLevel(level);
-				onConfigureScope(callback, scope);
+			scope.setLevel(level);
+			onConfigureScope(callback, scope);
```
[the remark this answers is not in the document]
tustanivsky: Yes, more of the same.

bitsandfoxes:
One more thing that comes to mind: We can mark those engine frames as `InAppExclude` so those won't get considered by the grouping algorithm. And those should be able to be propagated to the respective SDKs.

plugin-dev/Source/Sentry/Private/Android/SentrySubsystemAndroid.cpp
tustanivsky: That's a good point. I've moved the assertion-specific logic from `CaptureException` to `CaptureAssertion`. Also, in this case we can leave the `CaptureException` interface stub empty since on Android assert/ensure handling flow is different from other platforms.

tustanivsky:
Since there are quite a few questions regarding the assertions/ensures capturing on Android I'll try to address these with the following explanation:

The general idea of this PR is to hook into Unreal's error handling process using a custom `FSentryOutputDeviceError` so that whenever assertion is fired we capture a manually created exception event with a proper callstack and shut down the application.

The problem with this approach on Android is that Java event's callstack doesn't contain any C++ frames and thus it's not informative at all. There's no straightforward way to convert a native callstack which we can obtain on Unreal's side to a corresponding Java structure to override the default one either. However, instead of reporting assert in a described way within a single session we can set some flags identifying that it happened and force the app to crash. This allows to pass things over to `sentry-native` which captures a native crash with a "normal" C++ callstack. Now on the next application launch in `beforeSend` handler we verify if the crash was an Unreal assertion by checking aforementioned flags and cut off the redundant callstack part if there's any. Also, these flags (if any) should be removed and not get sent to Sentry since we need them only for client-side event pre-processing.

The assertion-specific flags are set to an event via tags and not context since `sentry-java` can't serialize the latter for native crashes and in this case the required values are lost upon the next application launch. I believe this is a known limitation of Java SDK which we've encountered a couple of times in the past.

plugin-dev/Source/Sentry/Private/Android/SentrySubsystemAndroid.cpp
tustanivsky: The problem with using context values here is that these are ignored by `sentry-native` when it captures a native crash on Android and thus there will be no way to identify whether the event is an assertion or not upon the next application launch.

plugin-dev/Source/Sentry/Private/Android/Java/SentryBridgeJava.java
tustanivsky: Do you mean moving `checkForUnrealException` to C++? If so, we'll probably have to introduce another wrapper for Java `SentryException` which I was trying to avoid.

plugin-dev/Source/Sentry/Private/Android/Java/SentryBridgeJava.java
tustanivsky: Yeah, the naming could be better here - changed it to a more generic `preProcessEvent` yet I'm open to considering other options.

bitsandfoxes (approved):

PR #544, opened 2024-04-22
Add ability to check if sentry should be initialised to prevent unnecessary warnings

Edstub207:
This was a change made by someone internally that's probs useful for others. 

We use it when setting up Sentry via Code instead than using the automatic logic.

### DIFF
--- a/plugin-dev/Source/Sentry/Public/SentrySubsystem.h
+++ b/plugin-dev/Source/Sentry/Public/SentrySubsystem.h
@@ -270,7 +270,7 @@ public:
 
 	bool IsSupportedForCurrentSettings()
 	{
-		return IsCurrentBuildConfigurationEnabled() && IsCurrentBuildTargetEnabled() && IsCurrentPlatformEnabled();
+		return IsCurrentBuildConfigurationEnabled() && IsCurrentBuildTargetEnabled() && IsCurrentPlatformEnabled() && EnableForPromotedBuildsOnly();
 	}
 private:
 	/** Adds default context data for all events captured by Sentry SDK. */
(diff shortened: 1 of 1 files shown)


bitsandfoxes:
Thanks for pushing this upstream!
Interestingly enough we have something similar in the Unity SDK named [`ShouldInitializeSdk`](https://github.com/getsentry/sentry-unity/blob/0609977715641e0bd1261a542dcb4efaf917a8a4/src/Sentry.Unity/SentryUnityOptionsExtensions.cs#L10C30-L10C49)

plugin-dev/Source/Sentry/Public/SentrySubsystem.h
```diff
@@ -268,6 +268,10 @@ class SENTRY_API USentrySubsystem : public UEngineSubsystem
 	UFUNCTION(BlueprintCallable, Category = "Sentry")
 	USentryTransaction* StartTransactionWithContextAndOptions(USentryTransactionContext* Context, const TMap<FString, FString>& Options);
 
+	bool IsSupportedForCurrentSettings()
+	{
+		return IsCurrentBuildConfigurationEnabled() && IsCurrentBuildTargetEnabled() && IsCurrentPlatformEnabled();
```
tustanivsky, line 273: I recall we've added an `EnableForPromotedBuildsOnly` flag along with these settings - shouldn't it also be checked here?

Edstub207: @tustanivsky this was updated btw.

tustanivsky: I think we don't have the `USentrySubsystem::EnableForPromotedBuildsOnly()` method at this moment - is this something specific to your fork repo?

tustanivsky: No worries, apparently I've misplaced my previous comment.

tustanivsky (approved):

PR #547, opened 2024-04-29
Fix build errors introduced after adding a check if event capturing is supported for current plugin settings

tustanivsky:
This PR fixes minor build errors introduced in #544

#skip-changelog


tustanivsky:
@Edstub207 Can you confirm whether these changes are still valid?

PR #559, opened 2024-05-15
Fix threading issues capturing log messages as breadcrumbs

tustanivsky:
This PR addresses an [issue](https://github.com/getsentry/sentry-unreal/issues/514#issuecomment-2077663102) which was discovered during the discussion around #514 

Since log messages, asserts, ensures, etc. can come from different threads adding breadcrumbs automatically on such events can lead to undefined results due to the following reasons:
- New breadcrumb `UObject` instance is created during garbage collection
- Concurrent access to internal `SentryScopeDesktop` breadcrumbs storage

In order to resolve this breadcrumbs can now be captured without instancing any extra objects and some threading considerations were added.

### DIFF
--- a/plugin-dev/Source/Sentry/Private/Android/SentrySubsystemAndroid.h
+++ b/plugin-dev/Source/Sentry/Private/Android/SentrySubsystemAndroid.h
@@ -12,6 +12,7 @@ public:
 	virtual bool IsEnabled() override;
 	virtual ESentryCrashedLastRun IsCrashedLastRun() override;
 	virtual void AddBreadcrumb(USentryBreadcrumb* breadcrumb) override;
+	virtual void AddBreadcrumbWithParams(const FString& Message, const FString& Category, const FString& Type, const TMap<FString, FString>& Data, ESentryLevel Level) override;
 	virtual void ClearBreadcrumbs() override;
 	virtual USentryId* CaptureMessage(const FString& message, ESentryLevel level) override;
 	virtual USentryId* CaptureMessageWithScope(const FString& message, const FConfigureScopeNativeDelegate& onConfigureScope, ESentryLevel level) override;
--- a/plugin-dev/Source/Sentry/Private/Desktop/SentryScopeDesktop.cpp
+++ b/plugin-dev/Source/Sentry/Private/Desktop/SentryScopeDesktop.cpp
@@ -279,4 +279,9 @@ void SentryScopeDesktop::Apply(USentryEvent* event)
 	}
 }
 
+void SentryScopeDesktop::AddBreadcrumb(TSharedPtr<SentryBreadcrumbDesktop> breadcrumb)
+{
+	BreadcrumbsDesktop.Add(breadcrumb);
+}
+
 #endif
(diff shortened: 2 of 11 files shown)


plugin-dev/Source/Sentry/Private/Desktop/SentryScopeDesktop.cpp
```diff
@@ -17,12 +17,26 @@ SentryScopeDesktop::SentryScopeDesktop()
...
+	Dist = Scope.Dist;
+	Environment = Scope.Environment;
+	FingerprintDesktop = Scope.FingerprintDesktop;
+	TagsDesktop = Scope.TagsDesktop;
+	ExtraDesktop = Scope.ExtraDesktop;
+	ContextsDesktop = Scope.ContextsDesktop;
+	BreadcrumbsDesktop = Scope.BreadcrumbsDesktop;
+	LevelDesktop = Scope.LevelDesktop;
+}
+
```
bitsandfoxes, line 31: Where is this coming from?

plugin-dev/Source/Sentry/Private/Desktop/SentryScopeDesktop.cpp
```diff
@@ -17,12 +17,26 @@ SentryScopeDesktop::SentryScopeDesktop()
...
 SentryScopeDesktop::~SentryScopeDesktop()
 {
 }
 
 void SentryScopeDesktop::AddBreadcrumb(USentryBreadcrumb* breadcrumb)
 {
+	FScopeLock Lock(&CriticalSection);
```
bitsandfoxes, line 38: That's neat, that it works like that.

plugin-dev/Source/Sentry/Private/Android/SentrySubsystemAndroid.h
```diff
@@ -12,6 +12,7 @@ class SentrySubsystemAndroid : public ISentrySubsystem
 	virtual bool IsEnabled() override;
 	virtual ESentryCrashedLastRun IsCrashedLastRun() override;
 	virtual void AddBreadcrumb(USentryBreadcrumb* breadcrumb) override;
```
bitsandfoxes, line 14: If we're no longer using the "plain" `AddBreadcrumb`, do we need it?

bitsandfoxes (approved):
Looks good!

plugin-dev/Source/Sentry/Private/Desktop/SentryScopeDesktop.cpp
tustanivsky: It's essential to implement a custom copy constructor for the scope class after we've introduced a critical section member.

Since `FCriticalSection` disables its default copy constructor using the default mechanism for scope isn't possible and results into compilation errors.

plugin-dev/Source/Sentry/Private/Android/SentrySubsystemAndroid.h
tustanivsky: It's still a part of the public API so I think we should leave it even though we're not using it internally anymore.

plugin-dev/Source/Sentry/Private/Desktop/SentryScopeDesktop.cpp
bitsandfoxes: TIL. Thanks!

PR #563, opened 2024-05-24
Add UE 5.4 support for plugin CI

tustanivsky:
In order to resolve issues with DDC generation due to the local ZenServer connection timeout IPv6 had to be enabled for the default docker container bridge network.

Also, since the new docker image is larger some extra space should be freed up on GitHub runner.


.github/workflows/ci.yml
```diff
@@ -183,6 +189,9 @@ jobs:
           docker exec --user root unreal bash -c "
             chown -R $uid /home/ue4/UnrealEngine/Engine/Binaries/ThirdParty/Mono/Linux ;
             chown -R $uid /home/ue4/UnrealEngine/Engine/Binaries/ThirdParty/DotNet "
+          # Remove corrupted engine files that have to be rebuild
```
bitsandfoxes, line 192: Why are there corrupted engine files?

tustanivsky: It's an engine bug yet I'll remove this for now since the fix is all about network setup.

.github/workflows/ci.yml
```diff
@@ -164,13 +168,15 @@ jobs:
           gid=1000     # the ue4 group in the docker container
           user='gh'
           set -x
+          docker network create --ipv6 --subnet 2001:0DB8::/112 ip6net
```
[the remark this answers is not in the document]
tustanivsky: Good point, I've added one for clarity. Also, see https://github.com/adamrehn/ue4-docker/issues/357 for more details

.github/workflows/ci.yml
```diff
@@ -145,10 +145,14 @@ jobs:
           sudo time rm -rf /usr/local/lib/android
           sudo time rm -rf /usr/share/dotnet
           sudo time rm -rf /usr/share/swift
+          sudo time rm -rf /usr/local/share/powershell
```
bitsandfoxes, line 148: We could reuse [this](https://github.com/getsentry/sentry-unity/blob/e1c1e4476e83212f0daec856dd018ec4ca71d3f4/.github/workflows/ci.yml#L273-L281) action from the Unity SDK CI that helps us clear out the runner for more space.

bitsandfoxes (approved):
Finally! Thanks for adding this!

.github/workflows/ci.yml
```diff
@@ -132,45 +132,38 @@ jobs:
...
-          sudo time rm -rf /var/lib/apt/lists/*
-          # time docker rmi $(docker image ls -aq)
-          # time du --max-depth=3 --threshold=100M -h /usr /opt /var 2>/dev/null | sort -hr
-          df -h
+      - name: Free Disk Space (Ubuntu)
+        uses: jlumbroso/free-disk-space@f68fdb76e2ea636224182cfb7377ff9a1708f9b8
+        with:
+          android: true
+          dotnet: true
+          haskell: true
```
bitsandfoxes, line 144: I think these could be removed. Those are opt-outs of the deletion.

.github/workflows/ci.yml
tustanivsky: Somehow this action still misses some bits that we used to remove manually and the UE 5.4 container can't start due to a lack of disk space.

PR #566, opened 2024-06-03
Fix crashes not being reported during garbage collection

### LINKED ISSUE #560: Garbage collection crash not reported
### Environment

How do you use Sentry?
Sentry SaaS (sentry.io)

Which version of the SDK?
[0.17.1](https://github.com/getsentry/sentry-unreal/releases/tag/0.17.1)

Git

Which version of Unreal?
5.3.2

Windows

### Steps to Reproduce

1. Trigger a crash during garbage collection

### Expected Result

Crashes should be reported.

### Actual Result

No crash is reported. The last line in the log is ``LogSentrySdk: invoking `on_crash` hook``.

What I think is happening here is that it fails on [this line in `SentrySubsystemDesktop::HandleBeforeSend()`](https://github.com/getsentry/sentry-unreal/blob/2fe8b5f8b5cfceac65155e9bcfbbb4d084a1de72/plugin-dev/Source/Sentry/Private/Desktop/SentrySubsystemDesktop.cpp#L62), because it is attempting to create a new UObject, but the engine does not allow you to create UObjects during garbage collection. Which means that any crashes that occur during GC would fail to send.

### Any logs or screenshots
This is the relevant tail of the log output.
```
[2024.05.14-18.40.10:855][412]LogOutputDevice: Warning: 

Script Stack (0 frames) :

[2024.05.14-18.40.10:855][412]LogWindows: Error: appError called: Fatal error: [File:D:\******\UE5\Engine\Source\Runtime\CoreUObject\Private\UObject\GarbageCollectionVerification.cpp] [Line: 155] 
Encountered 8 object(s) breaking Disregard for GC assumptions. Please check log for details.



[2024.05.14-18.40.10:855][412]LogWindows: Windows GetLastError: The operation completed successfully. (0)
[2024.05.14-18.40.10:855][412]LogSentrySdk: flushing session and queue before crashpad handler
[2024.05.14-18.40.10:855][412]LogSentrySdk: Verbose: invoking `on_crash` hook
```

tustanivsky:
Closes #560

### DIFF
--- a/plugin-dev/Source/Sentry/Private/Desktop/SentryEventDesktop.cpp
+++ b/plugin-dev/Source/Sentry/Private/Desktop/SentryEventDesktop.cpp
@@ -12,9 +12,10 @@ SentryEventDesktop::SentryEventDesktop()
 	EventDesktop = sentry_value_new_event();
 }
 
-SentryEventDesktop::SentryEventDesktop(sentry_value_t event)
+SentryEventDesktop::SentryEventDesktop(sentry_value_t event, bool isCrash)
 {
 	EventDesktop = event;
+	IsCrashEvent = isCrash;
 }
 
 SentryEventDesktop::~SentryEventDesktop()
@@ -56,7 +57,7 @@ ESentryLevel SentryEventDesktop::GetLevel() const
 
 bool SentryEventDesktop::IsCrash() const
 {
-	return false;
+	return IsCrashEvent;
 }
 
 #endif
--- a/plugin-dev/Source/Sentry/Private/Desktop/SentrySubsystemDesktop.cpp
+++ b/plugin-dev/Source/Sentry/Private/Desktop/SentrySubsystemDesktop.cpp
@@ -34,6 +34,7 @@
 #include "Runtime/Launch/Resources/Version.h"
 #include "GenericPlatform/GenericPlatformOutputDevices.h"
 #include "GenericPlatform/GenericPlatformCrashContext.h"
+#include "UObject/GarbageCollection.h"
 
 #if PLATFORM_WINDOWS
 #include "Windows/WindowsPlatformMisc.h"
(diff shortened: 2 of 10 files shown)


plugin-dev/Source/Sentry/Private/Desktop/SentryEventDesktop.cpp
```diff
@@ -56,7 +57,7 @@ ESentryLevel SentryEventDesktop::GetLevel() const
 
 bool SentryEventDesktop::IsCrash() const
 {
-	return false;
+	return IsCrashEvent;
```
bitsandfoxes, line 60: That is pretty huge! But I don't get why now? Should that have worked previously? I see it being set in the `HandleBeforeCrash` but it seems unrelated to the GC.

plugin-dev/Source/Sentry/Private/Desktop/SentrySubsystemDesktop.cpp
```diff
@@ -73,7 +80,26 @@ sentry_value_t HandleBeforeCrash(const sentry_ucontext_t *uctx, sentry_value_t e
...
+
+		return SentrySubsystem->GetBeforeSendHandler()->HandleBeforeSend(EventToProcess, nullptr)
+			? event
+			: sentry_value_new_null();
+	}
+	else
+	{
+		// If crash occured during garbage collection we can't just obtain a GC lock like with normal events
+		// since there is no guarantee it will be ever freed. In this case crash event will be reported
+		// without calling a `beforeSend` handler.
```
bitsandfoxes, line 100: That's a pretty huge gotcha that we should put on the docs.

bitsandfoxes (commented):
Looking good!

plugin-dev/Source/Sentry/Private/Desktop/SentryEventDesktop.cpp
tustanivsky: The fact that `IsCrash()` method for desktop events was simply a stub caught my eye only now while reworking `onCrash` handler. We don't really use its Win/Linux implementation within the SDK so I assume that's the reason why it remained unnoticed for so long.

Would you prefer to fix this in a separate PR or we're okay with merging it with other changes here?

plugin-dev/Source/Sentry/Private/Desktop/SentrySubsystemDesktop.cpp
tustanivsky: How about adding this to the "Known limitations" section in readme instead?

bitsandfoxes: Sounds good to me! Thanks!

bitsandfoxes (approved):

PR #570, opened 2024-06-05
Fix build errors when packaging plugin for iOS

tustanivsky:
#skip-changelog


bitsandfoxes (approved):
Looks good but why are those necessary? How'd you find them, during manual testing?

tustanivsky:
> Looks good but why are those necessary? How'd you find them, during manual testing?

Yes, I've noticed this problem while testing things after the most recent `sentry-cocoa` update in UE 5.4.

PR #581, opened 2024-06-18
Add API allowing to check if captured event is ANR error 

### LINKED ISSUE #574: Expose more sentry types to Unreal or expose an delegate for listening ANR

Currently, if I want to filter ANR in beforesend handler, I have to use multiple if-condition together like `!Event->IsCrash()` and `Event->GetLevel() == ESentryLevel::Error` and possibly `Event->GetMessage().StartWith(TEXT("App Hanging"))` to mimic the same conditions set in https://github.com/getsentry/sentry-cocoa/blob/05896990562960d04052bf5a1548e290d89228bf/Sources/Sentry/SentryANRTrackingIntegration.m#L87-L103 

And it's not accurate due to I don't have SentryException and SentryMechanism exposed to Unreal, nothing returns me a SentryANRExceptionType with a USentryEvent pointer.

Maybe you could expose those above,  OR give another handler specifically tailored with ANR event?

tustanivsky:
This PR adds a new method to `USentryEvent` which allows checking if the captured event is an ANR error. By default, ANR error tracking is turned off in plugin settings.

Note, that ANR error tracking works only on Android/Apple since `sentry-native` doesn't support it at the moment.

Closes #574


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Features
+
+- Add API allowing to check if captured event is ANR error ([#581](https://github.com/getsentry/sentry-unreal/pull/581))
```
bitsandfoxes, line 7:
```suggestion
- Added `IsAnr()` method to mobile `SentryEvents` to check if the event is caused by the Application Not Responding ([#581](https://github.com/getsentry/sentry-unreal/pull/581))
```

plugin-dev/Source/Sentry/Public/SentryEvent.h
```diff
@@ -48,6 +48,10 @@ class SENTRY_API USentryEvent : public UObject
 	UFUNCTION(BlueprintPure, Category = "Sentry")
 	bool IsCrash() const;
 
+	/** Gets flag indicating whether the event is an Application Not Responding (ANR) error. */
+	UFUNCTION(BlueprintPure, Category = "Sentry")
+	bool IsAppNotResponding() const;
```
bitsandfoxes, line 53: It looks like it's nice having it written out in the blueprints?

bitsandfoxes (approved):

plugin-dev/Source/Sentry/Public/SentryEvent.h
tustanivsky: We can add a `meta=(DisplayName="Is App Not Responding")` attribute for UFUNCTION here in order to display the full name in blueprints while still having `IsAnr` in C++ as everywhere else. Does that sound good?

bitsandfoxes: That'd be great!

Issue #505, opened 2024-03-01
SentryBeforeSendHandler prevents crash reporter from working on containerized Linux


etelyatn:
### Environment

How do you use Sentry?
Sentry SaaS (sentry.io) or self-hosted/on-premise (which version?)

Which version of the SDK?
0.15.1

How did you install the package?
Git

Which version of Unreal?
5.3.2

Is this happening in Unreal (editor) or on a player like Android, iOS, Windows?
Linux/Proton (Container), Pixel streaming

### Steps to Reproduce

1. Create a `MySentryBeforeSendHandler` and configure it
2. Make a crash
3. Crash reporter will not work on contenirized linux

On desktop Windows and Ubuntu seem to work well.
`CaptureEvent` also working well.

### Expected Result

Сrash reporter works correctly, sending a report.

### Actual Result

Nothing happens, the crash reporter is silent

1pietras:
Having similar issue, Sentry is not sending crash report on containerized Linux.
There are logs of trying to work on crash, after "Segmentation fault" there are no further logs, no crash report is sent

```
[2024.03.06-08.52.14:561][  0]LogSentrySdk: Sentry plugin auto initialization: false

[2024.03.06-08.52.14:639][  0]LogSentrySdk: using database path "/home/dsuser/Alara/.sentry-native"
[2024.03.06-08.52.14:639][  0]LogSentrySdk: starting transport
[2024.03.06-08.52.14:639][  0]LogSentrySdk: starting backend
[2024.03.06-08.52.14:639][  0]LogSentrySdk: starting crashpad backend with handler "/home/dsuser/Game/Plugins/Sentry/Binaries/Linux/crashpad_handler"
[2024.03.06-08.52.14:641][  0]LogSentrySdk: using minidump URL (...)
[2024.03.06-08.52.14:654][  0]LogSentrySdk: started crashpad client handler
[2024.03.06-08.52.14:655][  0]LogSentrySdk: processing and pruning old runs
[2024.03.06-08.52.14:656][  0]LogSentrySdk: Sentry initialization completed with result 0 (0 on success).

[2024.03.06-09.13.23:218][748]LogSentrySdk: flushing session and queue before crashpad handler
[2024.03.06-09.13.23:218][748]LogSentrySdk: invoking `on_crash` hook
Segmentation fault
```

Engine Version : 5.3.0
Plugin Version: 0.11.0
Installed by Git

tustanivsky:
[...]

PR #589, opened 2024-07-03
Fix calling `beforeSend` handler during post-loading

tustanivsky:
Related to #505 ([comment](https://github.com/getsentry/sentry-unreal/issues/505#issuecomment-2200566352))


plugin-dev/Source/Sentry/Private/Desktop/SentrySubsystemDesktop.cpp
```diff
@@ -102,9 +108,14 @@ sentry_value_t HandleBeforeCrash(const sentry_ucontext_t *uctx, sentry_value_t e
 		USentryEvent* EventToProcess = NewObject<USentryEvent>();
 		EventToProcess->InitWithNativeImpl(eventDesktop);
 
-		return SentrySubsystem->GetBeforeSendHandler()->HandleBeforeSend(EventToProcess, nullptr)
-			? event
-			: sentry_value_new_null();
+		USentryEvent* ProcessedEvent = EventToProcess;
+		if(!FUObjectThreadContext::Get().IsRoutingPostLoad)
+		{
+			// Executing UFUNCTION is allowed only when not post-loading
```
bitsandfoxes, line 114: Could you provide some context what this means?

tustanivsky: Post-loading is a part of an actor lifecycle that goes right after finishing loading it from disk (i.e. when a game map opened in editor). Typically at this point some related actor's versioning or fixup behavior should be called. It's unsafe to use any blueprint-implemented functions (like `HandleBeforeSend` in our case) during this stage so we've added an extra check here to avoid triggering a corresponding assertion.

bitsandfoxes (approved):

plugin-dev/Source/Sentry/Private/Desktop/SentrySubsystemDesktop.cpp
bitsandfoxes: Thanks for the explanation!

PR #594, opened 2024-07-09
Fix crash when re-initializing Sentry

### LINKED ISSUE #592: Crash when re-initializing the sentry subsystem
### Environment

How do you use Sentry?
Sentry SaaS (sentry.io)

Which version of the SDK?
Git-URL, 18.0 git hub release for unreal 5.3

Which version of Unreal?
Unreal 5.3.2

We only use sentry for the windows game target for now.

### Steps to Reproduce

1. Initialize the sentry subsystem with a call to InitWithSettings, Initialize or setting InitAutomatically in USentrySettings to true
2. Initialize the sentry subsystem again later when having more data by calling InitWithSettings or Initialize again without calling Deinitialize first
3. Wait for a log message to be printed by unreal

### Expected Result

The app should not crash.

### Actual Result

The app crashes

### Any logs or screenshots

The reason for the crash is that when calling Initialize multiple times the  `FSentryOutputDevice` instance in OutputDevice gets destroyed without deregistering them first from GLog (see [USentrySubsystem::ConfigureOutputDevice()](https://github.com/getsentry/sentry-unreal/blob/e68c009748a587420a9f111e2bd077bc70f75740/plugin-dev/Source/Sentry/Private/SentrySubsystem.cpp#L642) ). 

Similar problems might be in `USentrySubsystem::ConfigureOutputDeviceError()` as the old FSentryOutputDeviceError device is still in GError when it is torn down while being used at the same time as the parent for the newly created device.

tustanivsky:
This PR fixes the internal plugin crash when re-initializing `USentrySubsystem` without its explicit closing beforehand.

The crash was related to custom Sentry output devices that were not deregistered properly upon any subsequent initialization attempt.

Also, assertion/ensure delegates are now deregistered too whenever calling subsystem's `Close` method to avoid potential reporting of the same event multiple times if re-init occured.

Closes #592

### DIFF
--- a/plugin-dev/Source/Sentry/Private/SentrySubsystem.cpp
+++ b/plugin-dev/Source/Sentry/Private/SentrySubsystem.cpp
@@ -66,6 +66,18 @@ void USentrySubsystem::Deinitialize()
 
 void USentrySubsystem::Initialize()
 {
+	if (!SubsystemNativeImpl)
+	{
+		UE_LOG(LogSentrySdk, Warning, TEXT("Sentry subsystem is invalid and can't be initialized."));
+		return;
+	}
+
+	if (SubsystemNativeImpl->IsEnabled())
+	{
+		UE_LOG(LogSentrySdk, Warning, TEXT("Sentry is already initialized. It will be shut down automatically before re-init."));
+		Close();
+	}
+
 	const USentrySettings* Settings = FSentryModule::Get().GetSettings();
 
 	if(Settings->Dsn.IsEmpty())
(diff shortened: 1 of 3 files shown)


plugin-dev/Source/Sentry/Private/SentrySubsystem.cpp
```diff
@@ -154,11 +163,25 @@ void USentrySubsystem::Close()
...
+		}
+
 		GError = OutputDeviceError->GetParentDevice();
+		OutputDeviceError = nullptr;
+	}
+
+	if(OnEnsureDelegate.IsValid())
+	{
+		FCoreDelegates::OnHandleSystemEnsure.Remove(OnEnsureDelegate);
+		OnEnsureDelegate.Reset();
```
bitsandfoxes, line 184: Should this go into the closure above as well?

tustanivsky: No, in order to capture ensures we don't use any of our custom output devices but rather rely on the engine's `FCoreDelegates::OnHandleSystemEnsure` delegate.

bitsandfoxes (approved):
LGTM!

PR #604, opened 2024-07-30
Fix missing headers breaking non-unity builds in release 0.19.0.

Quogu:
When upgrading from release 0.16.0 to 0.19.0, our Unreal project failed to build due to missing headers. Adding references to both `HAL/FileManager.h` and `HAL/PlatformFileManager.h` fixed the build.


tustanivsky:
@Quogu Thank you for bringing this up! I've duplicated the changes you've introduced here in #606 and also tweaked our sample project build setting a bit to ensure that similar issues won't slip through our CI check again.

PR #606, opened 2024-07-31
Fix build errors due to missing includes

tustanivsky:
This PR adds missing includes that were causing build errors and disable Unity build and PCH filses usage for sample's editor target to detect these issues in CI automatically.

Related to #604


bitsandfoxes (approved):
Thanks for making CI more aware!

PR #612, opened 2024-08-12
Fix batch script 'exit' usage which breaks command chaining (&&).

webbju:
Refrain from using `exit` in Windows Batch scripts since this breaks behaviour in Unreal Engine that uses command chaining (`&&`) for `PostBuildStep` action commands ([`UEBuildTarget.cs:2864`](https://github.com/EpicGames/UnrealEngine/blob/release/Engine/Source/Programs/UnrealBuildTool/Configuration/UEBuildTarget.cs#L2864)). This means that the associated `.run` sentinel files aren't output, breaking minimal rebuild checks.

> Note 2: Command [EXIT](https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/exit_2) without parameter /B results always in exiting entire command process independent on calling hierarchy and independent on how the Windows command processor was started – with parameter /K to keep cmd.exe running as used when opening a command prompt window or with /C to close after command processing finished as used on double clicking a batch file. Therefore exit without /B should be used wisely in a batch file (best: never).[^1]

[^1]: https://stackoverflow.com/a/37518000

### DIFF
--- a/plugin-dev/Scripts/upload-debug-symbols-win.bat
+++ b/plugin-dev/Scripts/upload-debug-symbols-win.bat
@@ -17,7 +17,7 @@ echo Sentry: Start debug symbols upload
 
 if "%TargetType%"=="Editor" (
     echo Sentry: Automatic symbols upload is not required for Editor target. Skipping...
-    exit
+    exit /B 0
 )
 
 if "%TargetPlatform%"=="Win64" (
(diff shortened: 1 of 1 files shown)


tustanivsky (approved):

tustanivsky:
That's a great catch! Thank you @webbju

PR #616, opened 2024-08-19
Add setting that allows switching the internal Sentry database location on Windows/Linux

### LINKED ISSUE #614: should i gitignore .sentry-native?
I recently added the sentry plugin to my game, and seeing that `.sentry-native` kept changing, I added to gitignore.

Users have complained on different issues related to it. should I return it to the repo?

will it keep changing?

tustanivsky:
This PR adds a new plugin setting that allows users to select one of the pre-defined locations for storing the internal Sentry database used on Windows/Linux. Specifically, it is required to initialize the `sentry-native` library which UE plugin uses under the hood.

Having such flexibility allows to avoid potential issues with using Sentry in installed game builds that might be located in a write-protected folder (i.e. `Program Files` on Windows).

Closes #614


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Features
+
+- Add setting that allows switching the internal Sentry database location on Windows/Linux ([#616](https://github.com/getsentry/sentry-unreal/pull/616))
```
bitsandfoxes, line 7: Can we make the entry a bit more specific what options are available?
I.e. The config panel now supports the option to choose between X and Y?

tustanivsky: Good point, I've rephrased it to make things more explicit

plugin-dev/Source/Sentry/Private/Desktop/SentrySubsystemDesktop.cpp
```diff
@@ -154,6 +154,22 @@ void SentrySubsystemDesktop::InitWithSettings(const USentrySettings* settings, U
...
+	case ESentryDatabaseLocation::ProjectUserDirectory:
+		databaseParentPath = FPaths::ProjectUserDir();
+		break;
+	}
+
+	if(databaseParentPath.IsEmpty())
+	{
+		UE_LOG(LogSentrySdk, Warning, TEXT("Unknown Sentry database location. Falling back to FPaths::ProjectUserDir()."));
+		databaseParentPath = FPaths::ProjectUserDir();
+	}
```
bitsandfoxes, line 171: AFAIK, other SDKs would interpret a missing or invalid database path as having the feature disabled. But there it's most likely about event-offline-caching.
I guess sentry-native would not do much without it?

bitsandfoxes (approved):

plugin-dev/Source/Sentry/Private/Desktop/SentrySubsystemDesktop.cpp
tustanivsky: `sentry-native` database path will default to the current working directory if one is not provided explicitly so technically it should be able to continue working as usual. On the other hand, if it's invalid rather than missing native SDK initialization will fail. Anyway, having the above check in place should prevent us from getting into any of these situations.

PR #631, opened 2024-09-09
Add trace propagation

### LINKED ISSUE #627: Add trace propagation capabilities
The API did not exist when the Tracing API was first implemented.

- sentry-native [docs](https://docs.sentry.io/platforms/native/tracing/trace-propagation/custom-instrumentation/)
- sentry-cocoa [docs](https://docs.sentry.io/platforms/apple/tracing/trace-propagation/)
- sentry-java [docs](https://docs.sentry.io/platforms/java/tracing/trace-propagation/custom-instrumentation/)

tustanivsky:
This PR adds missing APIs required for the distributed tracing feature. Basically, now it's possible to create a new `TransactionContext` based on a given [trace header](https://develop.sentry.dev/sdk/telemetry/traces/#header-sentry-trace) and query existing span/transaction trace info for its further propagation.

Closes #627

### DIFF
--- a/plugin-dev/Source/Sentry/Private/Apple/SentryIdApple.cpp
+++ b/plugin-dev/Source/Sentry/Private/Apple/SentryIdApple.cpp
@@ -7,7 +7,12 @@
 
 SentryIdApple::SentryIdApple()
 {
+	// `SentryId` definition was moved to Swift so its name that can be recognized by UE should be taken from "Sentry-Swift.h" to successfully load class on Mac
+#if PLATFORM_MAC
+	IdApple = [[SENTRY_APPLE_CLASS(_TtC6Sentry8SentryId) alloc] init];
+#elif PLATFORM_IOS
 	IdApple = [[SENTRY_APPLE_CLASS(SentryId) alloc] init];
+#endif
 }
 
 SentryIdApple::SentryIdApple(SentryId* id)
--- a/plugin-dev/Source/Sentry/Private/Apple/SentrySubsystemApple.cpp
+++ b/plugin-dev/Source/Sentry/Private/Apple/SentrySubsystemApple.cpp
@@ -79,7 +79,6 @@ void SentrySubsystemApple::InitWithSettings(const USentrySettings* settings, USe
 				[options addInAppExclude:it->GetNSString()];
 			}
 			options.enableAppHangTracking = settings->EnableAppNotRespondingTracking;
-			options.enableTracing = settings->EnableTracing;
 			if(settings->EnableTracing && settings->SamplingType == ESentryTracesSamplingType::UniformSampleRate)
 			{
 				options.tracesSampleRate = [NSNumber numberWithFloat:settings->TracesSampleRate];
--- a/plugin-dev/Source/Sentry/Private/Apple/SentrySubsystemApple.cpp
+++ b/plugin-dev/Source/Sentry/Private/Apple/SentrySubsystemApple.cpp
@@ -319,9 +319,17 @@ USentryTransactionContext* SentrySubsystemApple::ContinueTrace(const FString& se
 	{
 		sampleDecision = traceParts[2].Equals(TEXT("1")) ? kSentrySampleDecisionYes : kSentrySampleDecisionNo;
 	}
-	
+
+	// `SentryId` definition was moved to Swift so its name that can be recognized by UE should be taken from "Sentry-Swift.h" to successfully load class on Mac
+
+#if PLATFORM_MAC
+	SentryId* traceId = [[SENTRY_APPLE_CLASS(_TtC6Sentry8SentryId) alloc] initWithUUIDString:traceParts[0].GetNSString()];
+#elif PLATFORM_IOS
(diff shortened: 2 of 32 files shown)


plugin-dev/Source/Sentry/Private/Apple/SentrySubsystemApple.cpp
```diff
@@ -79,7 +79,6 @@ void SentrySubsystemApple::InitWithSettings(const USentrySettings* settings, USe
 				[options addInAppExclude:it->GetNSString()];
 			}
 			options.enableAppHangTracking = settings->EnableAppNotRespondingTracking;
-			options.enableTracing = settings->EnableTracing;
```
bitsandfoxes, line 82: Why was that removed?

plugin-dev/Source/Sentry/Private/Apple/SentrySubsystemApple.cpp
```diff
@@ -303,3 +302,36 @@ USentryTransaction* SentrySubsystemApple::StartTransactionWithContextAndOptions(
 
 	return SentryConvertorsApple::SentryTransactionToUnreal(transaction);
 }
+
+USentryTransactionContext* SentrySubsystemApple::ContinueTrace(const FString& sentryTrace, const TArray<FString>& baggageHeaders)
```
bitsandfoxes, line 306: Can we add some tests for this?

plugin-dev/Source/Sentry/Private/Apple/SentryIdApple.cpp
```diff
@@ -7,7 +7,12 @@
 
 SentryIdApple::SentryIdApple()
 {
+	// `SentryId` definition was moved to Swift so its name that can be recognized by UE should be taken from "Sentry-Swift.h" to successfully load class on Mac
+#if PLATFORM_MAC
+	IdApple = [[SENTRY_APPLE_CLASS(_TtC6Sentry8SentryId) alloc] init];
```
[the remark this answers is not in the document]
tustanivsky: I think it's related to the way we're consuming `sentry-cocoa` on Mac - there it is necessary to load each symbol from the `sentry.dylib` [manually](https://github.com/getsentry/sentry-unreal/blob/595bf09c567561e2842c74976d61dd9d195f3ea9/plugin-dev/Source/Sentry/Private/Apple/Convenience/SentryMacro.h#L13) while on iOS all of the `Sentry.framework` classes are available right out of the box.

plugin-dev/Source/Sentry/Private/Desktop/SentrySubsystemDesktop.cpp
```diff
@@ -497,6 +497,21 @@ USentryTransaction* SentrySubsystemDesktop::StartTransactionWithContextAndOption
 	return StartTransactionWithContext(context);
 }
 
+USentryTransactionContext* SentrySubsystemDesktop::ContinueTrace(const FString& sentryTrace, const TArray<FString>& baggageHeaders)
+{
+	sentry_transaction_context_t* nativeTransactionContext = sentry_transaction_context_new("<unlabeled transaction>", "default");
```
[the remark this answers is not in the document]
tustanivsky: "unlabeled transaction" is a default name set to every [transaction context](https://github.com/getsentry/sentry-java/blob/8586d1fe888254e57186cfaa74266703741da4cf/sentry/src/main/java/io/sentry/TransactionContext.java#L132) that's created via the `continueTrace` method in `sentry-java`. 

Since neither `sentry-native` nor `sentry-cocoa` provide similar API or recommendations on how to handle naming in case of trace propagation I've followed the Java approach here to unify the behavior across all platforms.

plugin-dev/Source/Sentry/Private/Apple/SentrySubsystemApple.cpp
tustanivsky: `enableTracing` was deprecated and now providing one of the `sampleRate`/`samplingFunction` options explicitly is sufficient to let the SDK know that the tracing is enabled.

plugin-dev/Source/Sentry/Private/Apple/SentrySubsystemApple.cpp
tustanivsky: Yeah, I've added a basic one. The existing `sentry-native` API is quite limited though so it's somewhat problematic to set up thorough testing for the tracing feature.

plugin-dev/Source/Sentry/Private/Apple/SentryIdApple.cpp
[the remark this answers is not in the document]
tustanivsky: After investigating this a bit more it seems to be a common practice to use those mangled class names in a scenario like ours. Linking the [same thing](https://github.com/getsentry/sentry-dotnet/blob/1621b6c2828e414fcf7797b8acc5e32e91a1bc4a/src/Sentry.Bindings.Cocoa/ApiDefinitions.cs#L2437) in .NET SDK here just in case.

bitsandfoxes (approved):
Thanks!

Issue #502, opened 2024-02-22
sentry reported an error: Invalid org token (http status: 401) on Upload Debug Symbol Automatically when packaging for android


as3mbus:
### Environment

How do you use Sentry?
Sentry SaaS (sentry.io) or self-hosted/on-premise (which version?)

> Self Hosted

Which version of the SDK?
How did you install the package? (Git-URL, Assetstore)

> 0.15.1 Git Release - Marketplace Type

Which version of Unreal?

> 5.3.2

Is this happening in Unreal (editor) or on a player like Android, iOS, Windows?

> Unreal Editor when packaging for android

### Steps to Reproduce

1. configure sentry following getting started guide
2. set up Upload Debug Symbol Automatically
3. disabled android file server plugin
4. package android build for distribution with shipping build mode

### Expected Result

1. Build are produced without error
2. Debug Symbol are uploaded to sentry server

### Actual Result

1. Build Failed with sentry error saying  caused by: sentry reported an error: Invalid org token (http status: 401)
2. i've tried with windows build and it works just fine

### Any logs or screenshots

![image](https://github.com/getsentry/sentry-unreal/assets/16043293/970b7bb9-d7d3-45b9-bdb9-d61eec2005ae)

``` 
UATHelper: Packaging (Android (ASTC)): > Task :app:extractProguardFiles
UATHelper: Packaging (Android (ASTC)): > Task :app:processReleaseJavaRes NO-SOURCE
UATHelper: Packaging (Android (ASTC)): > Task :downloader_library:processReleaseJavaRes NO-SOURCE
UATHelper: Packaging (Android (ASTC)): > Task :downloader_library:bundleLibResRelease NO-SOURCE
UATHelper: Packaging (Android (ASTC)): > Task :permission_library:processReleaseJavaRes NO-SOURCE
UATHelper: Packaging (Android (ASTC)): > Task :permission_library:bundleLibResRelease NO-SOURCE
UATHelper: Packaging (Android (ASTC)): > Task :downloader_library:bundleLibRuntimeToJarRelease
UATHelper: Packaging (Android (ASTC)): > Task :permission_library:bundleLibRuntimeToJarRelease
UATHelper: Packaging (Android (ASTC)): > Task :downloader_library:bundleLibCompileToJarRelease
UATHelper: Packaging (Android (ASTC)): > Task :permission_library:bundleLibCompileToJarRelease
UATHelper: Packaging (Android (ASTC)): > Task :app:processReleaseResources
[...]

PR #633, opened 2024-09-10
Upgrade Android gradle plugin

tustanivsky:
This PR is supposed to address Android symbol upload issues that some Sentry self-hosted users faced previously due to the legacy gradle plugin being in use (i.e. #502).

In order to bump the Android gradle plugin version it was necessary to override engine's gradle version as well by replacing `gradle-wrapper.properties` file in build dir.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Add setting that allows switching between the project and user directories for the internal Sentry database location on Windows/Linux ([#616](https://github.com/getsentry/sentry-unreal/pull/616))
 - Add non-ASCII characters support for user messages ([#624](https://github.com/getsentry/sentry-unreal/pull/624))
 - Upgrade Android gradle plugin from v2.1.5 to v4.11.0 ([#633](https://github.com/getsentry/sentry-unreal/pull/633))
+- Add API to allow users to trace their distributed system and connect in-game with backend errors ([#631](https://github.com/getsentry/sentry-unreal/pull/631))
 - Allow overriding `UploadSymbolsAutomatically` via environment variable `SENTRY_UPLOAD_SYMBOLS_AUTOMATICALLY` ([#636](https://github.com/getsentry/sentry-unreal/pull/636))
 
 ### Fixes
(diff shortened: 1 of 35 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 - Add setting that allows switching between the project and user directories for the internal Sentry database location on Windows/Linux ([#616](https://github.com/getsentry/sentry-unreal/pull/616))
 - Add non-ASCII characters support for user messages ([#624](https://github.com/getsentry/sentry-unreal/pull/624))
+- Upgrade Android gradle plugin ([#633](https://github.com/getsentry/sentry-unreal/pull/633))
```
bitsandfoxes, line 9: A `from X to Y` would be nice. And maybe a line about what that change made work.

plugin-dev/Config/FilterPlugin.ini
```diff
@@ -7,10 +7,5 @@
 ;    /Extras/...
 ;    /Binaries/ThirdParty/*.dll
 /Scripts/*
-/LinuxIntermediate/Build/Linux/UnrealEditor/Inc/Sentry/UHT/*
-/LinuxIntermediate/Build/Linux/UnrealGame/Development/Sentry/*
-/LinuxIntermediate/Build/Linux/UnrealGame/Inc/Sentry/UHT/*
-/LinuxIntermediate/Build/Linux/UnrealGame/Shipping/Sentry/*
-/LinuxIntermediate/Build/Linux/x64/UnrealGame/Development/Sentry/*
-/LinuxIntermediate/Build/Linux/x64/UnrealGame/Shipping/Sentry/*
+/Gradle/*
```
bitsandfoxes, line 10: Is that change intentional? The how do the `LinuxIntermediate` related to `Gradle`?

plugin-dev/Source/Sentry/Sentry_Android_UPL.xml
```diff
@@ -137,7 +141,7 @@
     <buildscriptGradleAdditions>
         <insert>
             dependencies {
-                classpath 'com.android.tools.build:gradle:3.5.4'
+                classpath 'com.android.tools.build:gradle:7.4.2'
```
bitsandfoxes, line 144: That's the gradle tool itself, right? And that's a requirement for the sentry-android-gradle-plugin `4.11.0`?
Are we going to mess with user's build by messing with the gradle version that comes with the UE installation?

tustanivsky: Yes, the `io.sentry.android.gradle >= 3.0.0` requires `Android Gradle Plugin >= 7.0.0` according to docs.
I've tested this with a few different engine versions (including pretty old UE 4.27) and the build was successful however there's always a chance that some compatibility issues with 3rd-party plugins can arise.

plugin-dev/Config/FilterPlugin.ini
tustanivsky: Yes, aside from adding `Gradle` here I've also removed `LinuxIntermediate` stuff which is now redundant (probably I've missed deleting these entries earlier). Should we tackle this kind of clean-up in a separate PR?

plugin-dev/Source/Sentry/Sentry_Android_UPL.xml
```diff
@@ -134,22 +152,30 @@
...
+                        <log text="Using legacy Sentry Gradle plugin 2.1.5 for Android debug symbol upload."/>
+                        <insert>
+                            dependencies {
+                                classpath 'com.android.tools.build:gradle:3.5.4'
+                                classpath 'io.sentry:sentry-android-gradle-plugin:2.1.5'
+                            }
+                        </insert>
+                    </true>
+                    <false>
+                        <log text="Using Sentry Gradle plugin 4.11.0 for Android debug symbol upload."/>
```
bitsandfoxes, line 169: Is the hardcoded version number going to bite us in the butt later on?

bitsandfoxes (commented):
I'm not sure I see how this fixes the issue linked. The issue was resolved by adding the `defaults url`, right? How does the newer Android Gradle Plugin help with this? Is there anything else we're getting out of it?

The fallback is great, I really appreciate it, changing the user's Gradle version still makes me nervous.

tustanivsky:
> I'm not sure I see how this fixes the issue linked. The issue was resolved by adding the defaults url, right? How does the newer Android Gradle Plugin help with this? Is there anything else we're getting out of it?

With the newer Gradle plugin version adding `defaults.url` to `sentry.properties` when using Sentry self-hosted shouldn't be required any longer. It was a legacy issue but now the right URL can be extracted from the authentication token so users won't face the problem with uploading symbols.

plugin-dev/Source/Sentry/Sentry_Android_UPL.xml
tustanivsky: Since we don't have any means allowing to check how Sentry Gradle plugin performs in our CI probably it'd be safer to keep a hardcoded version here instead of bumping it automatically for the time being.

bitsandfoxes (approved):
Considering that we cannot offer ootb support for self-hosted and there is a fallback/opt-out for those that need it: LGTM! 🔥

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 - Add setting that allows switching between the project and user directories for the internal Sentry database location on Windows/Linux ([#616](https://github.com/getsentry/sentry-unreal/pull/616))
 - Add non-ASCII characters support for user messages ([#624](https://github.com/getsentry/sentry-unreal/pull/624))
+- Upgrade Android gradle plugin from v2.1.5 to v4.11.0 ([#633](https://github.com/getsentry/sentry-unreal/pull/633))
```
bruno-garcia, line 9: @tustanivsky @bitsandfoxes would this require a major bump of this SDK?

Or since we're still experimental (`0.x`) at least a callout in the changelog larger than this one liner?

tustanivsky: This isn't a breaking change so a major version bump is not required. However it'll make sense to mention in docs that there's a fallback allowing users to switch back to the legacy gradle plugin in case any issues occur.

PR #636, opened 2024-09-18
Allow override of UploadSymbolsAutomatically value from an environment variable.

Quogu:
The use-case for this is that we want to upload symbols in some CI builds, but not when developers build and cook locally, so currently we have to edit the config file DefaultConfig.ini as part of the CI process. This is cumbersome, as it has some awkward interactions with Perforce and e.g. TeamCity that make the state of the files on CI agents hard to reason about without adding a preamble script to every job that enforces the correct value, even ones that shouldn't need to touch this config file. It interacts especially poorly when a developer has edited DefaultConfig.ini in their Perforce changelist, and introduces risk when other jobs reusing the same workspace make commits back to Perforce.

Allowing an environment variable to be supplied instead will allow conditional enabling / disabling of symbol upload without having to edit files, making it much easier to implement without having to worry about the state of files on disk, so is much preferable.

I've tested this with Unreal Engine 5.4, and the environment variable is passed through from the call to UnrealAutomationTool through to invoking the post-build script.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Add setting that allows switching between the project and user directories for the internal Sentry database location on Windows/Linux ([#616](https://github.com/getsentry/sentry-unreal/pull/616))
 - Add non-ASCII characters support for user messages ([#624](https://github.com/getsentry/sentry-unreal/pull/624))
+- Allow overriding `UploadSymbolsAutomatically` via environment variable `SENTRY_UPLOAD_SYMBOLS_AUTOMATICALLY` ([#636](https://github.com/getsentry/sentry-unreal/pull/636))
 
 ### Fixes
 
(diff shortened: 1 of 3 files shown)


tustanivsky:
@Quogu Thank you for submitting this PR! The idea of configuring certain plugin behaviors via the environment variables has been there for a while so I guess symbol upload can become a good starting point for bringing it to life. We'll run some tests on our end and get back to you.

Meanwhile, would you mind adding a similar feature for Mac/Linux ([upload-debug-symbols.sh](https://github.com/getsentry/sentry-unreal/blob/main/plugin-dev/Scripts/upload-debug-symbols.sh)) to keep things consistent across the supported platforms?

Quogu:
Hi @tustanivsky, I've added an equivalent implementation in the bash script, and in isolation, the change appears to work in WSL bash, but I don't have the means of testing this in-situ and making sure that the environment variable actually comes through properly via UBT.

tustanivsky (approved):

tustanivsky:
Looks good, thanks again @Quogu!

I'll make sure this is also mentioned in [docs](https://docs.sentry.io/platforms/unreal/configuration/debug-symbols/).

PR #646, opened 2024-10-07
Clean up symbol upload script logs

tustanivsky:
Related to #636 

#skip-changelog

### DIFF
--- a/plugin-dev/Scripts/upload-debug-symbols-win.bat
+++ b/plugin-dev/Scripts/upload-debug-symbols-win.bat
@@ -33,11 +33,10 @@ if "%TargetPlatform%"=="Win64" (
 )
 
 call :ParseIniFile "%ConfigPath%\DefaultEngine.ini" /Script/Sentry.SentrySettings UploadSymbolsAutomatically UploadSymbols
-echo Sentry: Automatic symbols upload is '%UploadSymbols%' from plugin settings
 
 if /i "%SENTRY_UPLOAD_SYMBOLS_AUTOMATICALLY%" NEQ "" (
   set UploadSymbols=%SENTRY_UPLOAD_SYMBOLS_AUTOMATICALLY%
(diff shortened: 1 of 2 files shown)


plugin-dev/Scripts/upload-debug-symbols-win.bat
```diff
@@ -33,11 +33,10 @@ if "%TargetPlatform%"=="Win64" (
 )
 
 call :ParseIniFile "%ConfigPath%\DefaultEngine.ini" /Script/Sentry.SentrySettings UploadSymbolsAutomatically UploadSymbols
-echo Sentry: Automatic symbols upload is '%UploadSymbols%' from plugin settings
```
bitsandfoxes, line 36: Won't we miss that log?

plugin-dev/Scripts/upload-debug-symbols.sh
```diff
@@ -31,11 +31,10 @@ else
 fi
 
 UPLOAD_SYMBOLS=$(awk -F "=" '/UploadSymbolsAutomatically/ {print $2}' "${CONFIG_PATH}/DefaultEngine.ini")
-echo "Sentry: Automatic symbols upload is '$UPLOAD_SYMBOLS' from plugin settings"
```
bitsandfoxes, line 34: Won't we miss that log?

bitsandfoxes (commented):
Improved logs! <3

plugin-dev/Scripts/upload-debug-symbols-win.bat
tustanivsky: If symbol upload flag is unspecified/disabled in settings or overridden by env variable we'll print a corresponding message later on anyway, so keeping this message here doesn't seem to be necessary any longer.

bitsandfoxes (approved):

Issue #645, opened 2024-10-04
Upload Android NDK native symbols when packaging


bruno-garcia:
Are we building Android from source? if so, we have to upload the native symbols to Sentry.

I noticed it's commented out now: https://github.com/getsentry/sentry-unreal/blob/main/.craft.yml#L5-L9

PR #649, opened 2024-10-08
upload symbols for NDK builds

bruno-garcia:
Resolves https://github.com/getsentry/sentry-unreal/issues/645


bitsandfoxes (approved):
Go upload! 🎉

Issue #575, opened 2024-06-10
Improve CI - test on more versions, install final package for validation, test on more linux distros


bitsandfoxes:
The ideal state of CI would be the following:
- Install the package on Windows, Linux and macOS
- Target supported Unreal versions
  - Run CI on the oldest, supported version: https://github.com/getsentry/sentry-unreal/issues/661 (we might want to check with Data whether this is an issue in the first place but based on the inquiries we're getting it does not seem we can skip 4.27)
  - Prevent dependency related regressions. I.e. run on Amazon Linux https://github.com/getsentry/sentry-unreal/issues/635
- Run 
- Build a sample project on those platforms for all supported target platforms: Android, iOS, Windows, Linux, and macOS
- Run those artifacts as integration/smoke test https://github.com/getsentry/sentry-unreal/issues/77

bitsandfoxes:
This can be broken down, prioritizing Windows/Linux setup and builds.

bruno-garcia:
> Prevent dependency related regressions. I.e. run on Amazon Linux https://github.com/getsentry/sentry-unreal/issues/635

Can this be done by creating docker images using different distros as the base OS?

tustanivsky:
> Can this be done by creating docker images using different distros as the base OS?

I've followed this approach when testing Amazon Linux thing locally so there shouldn't be any blockers for doing that in CI.

PR #658, opened 2024-10-14
ci: heavy linux workload on ubuntu-latest-4-cores

bruno-garcia:
Also fixes https://github.com/getsentry/sentry-unreal/issues/473#issuecomment-2398285546

image

Looking at the large runner [here](https://github.com/getsentry/sentry-dotnet/actions/runners), looks like we need to use the right name

#skip-changelog

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,7 @@ jobs:
     uses: ./.github/workflows/sdk.yml
     with:
       target: Android
-      runsOn: ubuntu-latest
+      runsOn: ubuntu-latest-4-cores
 
   ios-sdk:
     uses: ./.github/workflows/sdk.yml
@@ -44,7 +44,7 @@ jobs:
   build:
     needs: [android-sdk, ios-sdk, macos-sdk, linux-sdk, windows-crashpad-sdk, windows-breakpad-sdk]
     name: Package
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-latest-4-cores
     steps:
       - name: Checkout
         uses: actions/checkout@v3
@@ -126,7 +126,7 @@ jobs:
 
   test:
     name: Test UE ${{ matrix.unreal }}
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-latest-4-cores
 
     strategy:
       fail-fast: false
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -180,7 +180,7 @@ jobs:
             ghcr.io/epicgames/unreal-engine:dev-slim-${{ matrix.unreal }}.1
           docker logout ghcr.io
           # Add the user so it has a home directory (needed for the pip cache later on)
-          docker exec --user root unreal useradd --non-unique -u $uid -g $gid --create-home $user
+          docker exec --user root unreal useradd -u $uid -g $gid --create-home $user
           # Ensure CA certs are in the right directory (needed for running tests)
           docker exec --user root unreal bash -c "
             mkdir -p /etc/pki/tls/certs ;
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -160,17 +160,6 @@ jobs:
           # time du --max-depth=3 --threshold=100M -h /usr /opt /var 2>/dev/null | sort -hr
           df -h
 
-      - name: Modify current user UID
-        run: |
-          echo "Current user and UID:"
-          id          
-          echo "Modifying UID..."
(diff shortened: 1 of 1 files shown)


bruno-garcia:
A problem I see already is that it'll run the build in series. Since we have 1 agent, and 3 parallel builds for it.

bruno-garcia:
A problem I see already is that it'll run the build in series. Since we have 1 agent, and 3 parallel builds for it.

tustanivsky:
> ```
> Status: Downloaded newer image for ghcr.io/epicgames/unreal-engine:dev-slim-5.4.1
> 29d835b3ddc370bc69d060b9993670e0e890fc1a2313a49416fe5c09f66844c8
> + docker logout ghcr.io
> Removing login credentials for ghcr.io
> + docker exec --user root unreal useradd -u 1000 -g 1000 --create-home gh
> useradd: UID 1000 is not unique
> Error: Process completed with exit code 4.
> ```
> 
> @tustanivsky any idea what's up?

After investigating this for a while I've figured out that there are some differences between the standard and large GitHub runners in terms of user configuration. On both of them we run [agent](https://github.com/actions/runner) on behalf of `runner` user however it has different UIDs:

- `1001` on standard runner:
```
$ cat /etc/passwd
...
runneradmin:x:1000:1000:Ubuntu:/home/runneradmin:/bin/bash
runner:x:1001:122:,,,:/home/runner:/bin/bash
```
- `1000` on large runner:
```
$ cat /etc/passwd
...
runner:x:1000:1000:Ubuntu:/home/runner:/bin/bash
```

At the same time UE docker container provided by Epic Games already has a `ue4` user with UID `1000` which causes the above error when we're running the following command:

https://github.com/getsentry/sentry-unreal/blob/311b0058321ff048ed2b13039d24586b5400bb5e/.github/workflows/ci.yml#L183

This step is required to install [UE4 CLI](https://github.com/adamrehn/ue4cli) that we're using later on to build the plugin and run tests. Since it's just a convenience utility for us I'll see if we can get rid of it and use the Unreal's build tools directly to avoid dealing with that matching UIDs issue.

.github/workflows/ci.yml
```diff
@@ -126,7 +126,7 @@ jobs:
 
   test:
     name: Test UE ${{ matrix.unreal }}
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-latest-4-cores-arm64
```
bruno-garcia, line 129: @bitsandfoxes should we make this a matrix instead? So it runs on both 4-cores and 4-cores-arm64?

bruno-garcia: Oh actually that won't work because there are not arm64 docker images for UE?

> WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested

tustanivsky (approved):
Removing `ue4cli` usages from the existing workflow spared us from creating additional user in docker container and allowed to avoid UIDs collision problem entirely.

Also, with this large GitHub runner we now can package the sample app in CI for all three UE version within a reasonable amount of time and without a risk to run out of space.

.github/workflows/ci.yml
tustanivsky: There's no official arm64 docker images for UE so it won't work (for now at least). However we may utilize our new arm64-runner for the future CI pipeline improvements (#575) and proper support of this arch (#473)

tustanivsky:
@bruno-garcia Are there any additional changes required within the scope of this PR beyond the introduction of the `ubuntu-latest-4-cores` runner to our CI pipeline?

.github/workflows/ci.yml
```diff
@@ -221,16 +213,28 @@ jobs:
...
+          docker exec -w /workspace/checkout/${{ matrix.app }} unreal /home/ue4/UnrealEngine/Engine/Binaries/Linux/UnrealEditor \
             /workspace/checkout/${{ matrix.app }}/SentryPlayground.uproject \
             -ReportExportPath=/workspace/checkout/${{ matrix.app }}/Saved/Automation \
             -ExecCmds="Automation RunTests Sentry;quit" \
             -TestExit="Automation Test Queue Empty" \
             -Unattended \
             -NoPause \
             -NoSplash \
-            -NullRHI
+            -NullRHI
```
bitsandfoxes, line 237: Could you explain what's going on here? If I read that right:
1. We're first building the project via the `RunUAT.sh` but with **all** the flags.
2. We're copying some logs to somewhere else
3. We're running some tests but from within the Editor
Do I get that right?

bitsandfoxes (approved):

Issue #664, opened 2024-10-23
Update Native SDK


tustanivsky:
Building the Sentry Native SDK is an essential part of Unreal Engine plugin CI workflow which is required to support automatic crash capturing on Linux.

Starting from Native SDK version [0.7.7](https://github.com/getsentry/sentry-native/releases/tag/0.7.7) we've faced issues with updating the `sentry-native` submodule caused by multiple build errors in CI ([workflow file](https://github.com/getsentry/sentry-unreal/blob/main/.github/workflows/sdk.yml), [build script](https://github.com/getsentry/sentry-unreal/blob/main/scripts/build-linux.sh)). Specifically, the upgrade of `mini_chromium` (one of the crashpad's dependencies) now requires C++20 and that causes compatibility issues with older [toolchains used in Unreal Engine](https://dev.epicgames.com/documentation/en-us/unreal-engine/linux-development-requirements-for-unreal-engine?application_version=5.5) and our CI environment.

The attempt to address the above problem was made in scope of #639. Switching to newer `clang-13` and `libc++-16` in CI fixed the Native SDK build and even allowed us to pass the test checks for UE 5.3/5.4 however it still fails for UE 5.2 (and probably for older versions too) due to mismatched versions of `GLIBC` and `GLIBCXX`. This could mean that Docker containers provided by Epic Games which we're using for testing don’t meet Unreal Engine's toolchain requirements. We've already been in a somewhat similar situation in #173 that made us stick to the older `Ubuntu-20.04` version to produce Native SDK build artifacts.

Possible alternative workarounds:
- Consider using `breakpad` instead of `crashpad` (not ideal)
- Investigate if there's a proper way to hook into the engine's error-handling flow to send crashes to Sentry

PR in UE SDK which's also related to this subject: https://github.com/getsentry/sentry-unreal/pull/657

Related issue in Native SDK: https://github.com/getsentry/sentry-native/issues/1041.

bruno-garcia:
> Investigate if there's a proper way to hook into the engine's error-handling flow to send crashes to Sentry

Agree this is worth looking at. If anything we'll have the same level of support for crash handling than folks are used to with the built-in crash reporter. That

>  even allowed us to pass the test checks for UE 5.3/5.4 however it still fails for UE 5.2 

[...]

PR #675, opened 2024-11-01
chore: Add UE 4.27, 5.0 and 5.1 to CI pipeline

### LINKED ISSUE #661: Add UE4.27 to CI
To make sure we stay backwards compatible with the last version of UE4.

To avoid regressions, like this one fixed recently:
* https://github.com/getsentry/sentry-unreal/pull/660

tustanivsky:
This PR ensures that the plugin stays backward compatible with some older Unreal versions (including UE 4.27) and addresses minor build errors that were revealed during testing.

The plugin will continue to support at least the latest three Unreal Engine versions to comply with Marketplace guidelines while the decision to drop support for older ones may be influenced by the outcome of #664.

It's also worth noticing that the corresponding packages will be added to the release page.

Closes #661 
Related to #575


plugin-dev/Source/Sentry/Private/Desktop/SentrySubsystemDesktop.cpp
```diff
@@ -74,7 +74,7 @@ void PrintVerboseLog(sentry_level_t level, const char *message, va_list args, vo
 
 void PrintCrashLog(const sentry_ucontext_t *uctx)
 {
-#if PLATFORM_WINDOWS
+#if PLATFORM_WINDOWS && ENGINE_MAJOR_VERSION == 5
```
bruno-garcia, line 77: How about this so if/when 6 lands, it's included automatically? (I assume nothign would work on UE3 anyway)?

```suggestion
#if PLATFORM_WINDOWS && ENGINE_MAJOR_VERSION != 4
```

bruno-garcia (approved):
1 suggestion on how to deal with ifdef and verison, but that thought process would need to be applied everywhere

plugin-dev/Source/Sentry/Private/Desktop/SentrySubsystemDesktop.cpp
tustanivsky: Since we have quite a few places with the version-specific code I've created a separate issue to track this request - #676

PR #681, opened 2024-11-08
chore: Download the `Cocoa SDK` instead of building it

### LINKED ISSUE #671: Reduce complexity of maintaining the `Cocoa SDK` plugin dependency
Building `sentry-cocoa` is currently part of the UE plugin [CI workflow](https://github.com/getsentry/sentry-unreal/blob/main/.github/workflows/ci.yml). In case of significant changes, regular updates of the corresponding submodule require additional effort to fix the existing build scripts for [iOS](https://github.com/getsentry/sentry-unreal/blob/main/scripts/build-ios.sh) and [Mac](https://github.com/getsentry/sentry-unreal/blob/main/scripts/build-mac.sh).

There are a few options to consider that can help to reduce the complexity of maintaining `Cocoa SDK` dependency:
- Download the required `sentry-cocoa` binaries from its GitHub release page ([example](https://github.com/getsentry/sentry-unity/pull/1854) of how this was done in Unity)
- Reuse the existing [build-xcframework.sh](https://github.com/getsentry/sentry-cocoa/blob/main/scripts/build-xcframework.sh) script which  `sentry-cocoa` relies on itself. Its current version is not optimized for Unreal (i.e. UE doesn't support Mac Catalyst or iPhone simulator) however it should be possible to introduce an extra parameter that will reduce build times and allow to skip unnecessary steps.

tustanivsky:
Closes #671 

#skip-changelog


bitsandfoxes (commented):
Why are you splitting downloading iOS and macOS? It looks like you're taking them from the same downloaded xcframework?

bitsandfoxes:
Could that not be a "prepare Cocoa" script/step? Instead of splitting it in two?

tustanivsky:
I've merged `download-ios.sh` and `download-mac.sh` so that all cocoa-related preparations happen within a single script.

.github/workflows/update-dependencies.yml
```diff
@@ -29,8 +29,8 @@ jobs:
   cocoa:
     uses: getsentry/github-workflows/.github/workflows/updater.yml@v2
     with:
-      path: modules/sentry-cocoa
-      name: Cocoa SDK (iOS)
+      path: modules/sentry-cocoa.properties
+      name: Cocoa SDK (iOS and Mac)
```
bruno-garcia, line 33: Don't we use sentry-native on Mac?

CHANGELOG.md
```diff
@@ -6,30 +6,22 @@
 
 - Assertions now print their callstack a log file ([#637](https://github.com/getsentry/sentry-unreal/pull/637))
 - Add UE 4.27, 5.0 and 5.1 to CI pipeline ([#675](https://github.com/getsentry/sentry-unreal/pull/675))
+- Download the `Cocoa SDK` instead of building it ([#681](https://github.com/getsentry/sentry-unreal/pull/681))
```
bruno-garcia, line 9: I don't think this needs to be in the changelog since for the end user, it doesn't make a difference, does it?

tustanivsky: You're right, this can be removed.

.github/workflows/update-dependencies.yml
tustanivsky: No, we switched to using sentry-cocoa on Mac quite a while ago (#190).

bruno-garcia (approved):

PR #688, opened 2024-11-18
Fix issue with `MaxBreadcrumbs` setting is not respected on desktop

### LINKED ISSUE #683: MaxBreadcrumbs setting is not respected and leads to failures to ingest events
### Environment

How do you use Sentry?
Sentry SaaS (sentry.io) or self-hosted/on-premise (which version?)

Which version of the SDK?

0.18

git

Which version of Unreal?

5.4.4

All desktop platforms at least

### Steps to Reproduce

1. Call SentrySubsystem::AddBreadcrumb or SentrySubsystem::AddBreadcrumbWithParams more times than the MaxBreadcrumbs config, which is 100 by default
2. Send a Sentry event

### Expected Result

I'd expect the breadcrumbs to truncate or include the most recent MaxBreadcrumbs breadcrumbs instead of growing a TArray and including the whole thing indefinitely.

### Actual Result

All of the breadcrumbs added, including those over the limit, are included in the event or the event is completely rejected for being too large

### Any logs or screenshots

tustanivsky:
Since `sentry-native` doesn't support working with multiple scopes we've added a custom scope class implementation for Windows/Linux to align the UE plugin API across different platforms (see #328). `SentryScopeDesktop` stores breadcrumbs in its own cache and thus ignores `MaxBreadcrumbs` setting value provided during the Native SDK initialization. This PR adds a corresponding breadcrumbs amount check and changes the cache's container type to ringbuffer.

Closes #683

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@
 - Add LinuxArm64 support for UE plugin ([#672](https://github.com/getsentry/sentry-unreal/pull/672))
 - Add UE 4.27, 5.0 and 5.1 to CI pipeline ([#675](https://github.com/getsentry/sentry-unreal/pull/675))
 
+### Fixes
+
+- Fix issue with `MaxBreadcrumbs` setting is not respected on desktop ([#688](https://github.com/getsentry/sentry-unreal/pull/688))
+
 ### Dependencies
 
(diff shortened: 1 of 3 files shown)


bruno-garcia (approved):

bruno-garcia:
@tustanivsky approved to unblock. We need some tests though

PR #696, opened 2024-11-22
Fix crash on Android when starting/ending session manually

### LINKED ISSUE #694: USentrySubsystem::StartSession() crashes games on the Android platform
### Environment

**How do you use Sentry?**
Sentry SaaS (sentry.io)

**Which version of the SDK?**
0.21.0

**How did you install the package? (Git-URL, Assetstore)**
Using the crash reporter

**Which version of Unreal?**
5.3

### Problem

The calling of the method `USentrySubsystem::StartSession()` crashes games on the Android platform. 

Under the hood it calls `SentrySubsystemAndroid::StartSession`, which calls `SentryBridgeJava::startSession`. However,  `io.sentry.unreal.SentryBridgeJava` defined [here](https://github.com/getsentry/sentry-unreal/blob/0.21.0/plugin-dev/Source/Sentry/Private/Android/Java/SentryBridgeJava.java) doesn't have such a method. 

As a result this check fails:
```
checkf(Method.Method, TEXT("Unable to find Java Method %s with Signature %s"), UTF8_TO_TCHAR(MethodName), UTF8_TO_TCHAR(FunctionSignature));
```
in method `FSentryJavaObjectWrapper::GetStaticMethod`.

`USentrySubsystem::StartSession` is being called after application started inside `AGameMode::StartPlay`.

The reason is the same method is being called from here:
https://github.com/getsentry/sentry-unreal/blob/0.21.0/plugin-dev/Source/Sentry/Private/Android/SentrySubsystemAndroid.cpp#L257

But as stated, there is no definition of the method in code here:
https://github.com/getsentry/sentry-unreal/blob/0.21.0/plugin-dev/Source/Sentry/Private/Android/Java/SentryBridgeJava.java

tustanivsky:
This PR sets the right class name for native `startSession`/`endSession` methods invocation on Android.

Closes #694

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixes
+
+- Fix crash on Android when starting/ending session manually ([#696](https://github.com/getsentry/sentry-unreal/pull/696))
+
 ### Dependencies
 
 - Bump Native SDK from v0.7.12 to v0.7.15 ([#690](https://github.com/getsentry/sentry-unreal/pull/690), [#693](https://github.com/getsentry/sentry-unreal/pull/693), [#695](https://github.com/getsentry/sentry-unreal/pull/695))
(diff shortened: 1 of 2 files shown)


PlasmaDev5 (approved):
Simple fix. Approved merge

Issue #496, opened 2024-02-15
Sessions not getting marked as crashed


gid-sentry:
### Environment

How do you use Sentry?
Sentry SaaS (sentry.io) 

Which version of the SDK? 0.15.1

How did you install the package? AssetStore

### Steps to Reproduce

1. Record a Session
2. Crash it

### Expected Result

Session appears in Sentry as crashed

### Actual Result

Session appears in Sentry, but not as crashed. Crash makes it into Sentry though .

>So I've tried setting this value Settings->EnableAutoSessionTracking = true; in USentrySettings . That didnt seem to do anythihng
And then I tried calling USentrySubsystem::StartSession();, which then does show up  as a session in the dashboard. But then when I attempt to crash from that session, it still counts as successful. I made sure not to call USentrySubsystem::EndSession()
And then I tried "hack" the Sentry library code in SentrySubsystemDesktop::HandleBeforeSend code to call sentry_end_session_with_status with a crash status defined in sentry.h. This didnt seem to work either because I think HandleBeforeSend doesnt seem to be called..

gid-sentry:
Internal ticket here: https://getsentry.atlassian.net/browse/FEEDBACK-2108

bruno-garcia:
@tustanivsky any idea what might be happening?

bruno-garcia:
@tustanivsky is that somewhere on our docs? If not, we should add it

strangezakary:
Hi we are experiencing this issue as well. We are on Unreal 5.3 and are getting errors with crash_type: crash and we are still seeing 100% crash-free. We are using automatic session tracking, any ideas what to look at?

strangezakary:
Looking more into this it doesn't look like sentry_end_session_with_status is called anywhere so how would it know if it crashed?

tustanivsky:
@strangezakary Do you have `Enable automatic crash capturing (Windows, UE 5.2+)` option turned on in Sentry plugins settings?

strangezakary:
@tustanivsky We do. I think the default is on

strangezakary:
Actually it looks like we don't? Do we need this on?

tustanivsky:
Yes, this has to be enabled. Basically, it allows you to utilize Sentry's native integration for capturing crashes instead of the engine's default error-handling mechanism and thus track session status properly.

tustanivsky:
[...]

Issue #381, opened 2023-09-13
Support non-crashing aborts


kristjanvalur:
UnrealEngine sometimes decides to exit with an error, even without crashing.  Such crashes are not caught by Sentry.  We have been using our own plugin for years, (https://github.com/mainframeindustries/UE4SentryClientPlugin) but decided to move to this _standard_ one, also to try to address this issue.

The problem is probably not currently solvable, but might be, with some work.

The issue is code like this:
```
// Check for GPU completion
			uint64 CompletedFenceValue = CurrentQueue.Fence.D3DFence->GetCompletedValue();

			if (CompletedFenceValue == UINT64_MAX)
			{
				// If the GPU crashes or hangs, the driver will signal all fences to UINT64_MAX. If we get an error code here, we can't pass it directly to 
				// VERIFYD3D12RESULT, because that expects DXGI_ERROR_DEVICE_REMOVED, DXGI_ERROR_DEVICE_RESET etc. and wants to obtain the reason code itself
				// by calling GetDeviceRemovedReason (again).
				HRESULT DeviceRemovedReason = CurrentQueue.Device->GetDevice()->GetDeviceRemovedReason();
				if (DeviceRemovedReason != S_OK)
				{
					VerifyD3D12Result(DXGI_ERROR_DEVICE_REMOVED, "CurrentQueue.Fence.D3DFence->GetCompletedValue()", __FILE__, __LINE__, CurrentQueue.Device->GetDevice());
				}
			}
```

The `VerifyD3D12Result` will, if failing, perform this:

```
	void VerifyD3D12Result(HRESULT D3DResult, const ANSICHAR* Code, const ANSICHAR* Filename, uint32 Line, ID3D12Device* Device, FString Message)
	{
		check(FAILED(D3DResult));
		
		GD3DCallFailedCS.Lock();

		const FString& ErrorString = GetD3D12ErrorString(D3DResult, Device);
		UE_LOG(LogD3D12RHI, Error, TEXT("%s failed \n at %s:%u \n with error %s\n%s"), ANSI_TO_TCHAR(Code), ANSI_TO_TCHAR(Filename), Line, *ErrorString, *Message);
		
		if (D3DResult == E_OUTOFMEMORY)
		{
			TerminateOnOutOfMemory(Device, D3DResult, false);
		}
		else
		{
			TerminateOnGPUCrash(Device, nullptr, 0);
		}

		// Make sure the log is flushed!
		GLog->Panic();

[...]

PR #697, opened 2024-11-22
Add Command Line Arguments to DefaultContext

jayhf:
The command line arguments seem like a really useful piece of information to have when debugging crashes. This adds them to the DefaultContext so that they're readily available in Sentry

### DIFF
--- a/plugin-dev/Source/Sentry/Private/SentrySubsystem.cpp
+++ b/plugin-dev/Source/Sentry/Private/SentrySubsystem.cpp
@@ -489,6 +489,7 @@ void USentrySubsystem::AddDefaultContext()
 	DefaultContext.Add(TEXT("Is standalone"), FApp::IsStandalone() ? TEXT("True") : TEXT("False"));
 	DefaultContext.Add(TEXT("Is unattended"), FApp::IsUnattended() ? TEXT("True") : TEXT("False"));
 	DefaultContext.Add(TEXT("Game name"), FApp::GetName());
+	DefaultContext.Add(TEXT("Command Line Arguments"), FCommandLine::GetOriginal());
 
 	SubsystemNativeImpl->SetContext(TEXT("Unreal Engine"), DefaultContext);
 }


tustanivsky:
Hey @jayhf, I believe we're already adding the command line input to `Crash Info` context on Win/Linux for captured crash events:

https://github.com/getsentry/sentry-unreal/blob/982894bd75f914789ef9e9c14aaab5d774b71324/plugin-dev/Source/Sentry/Private/Desktop/CrashReporter/SentryCrashContext.cpp#L54

Does it miss some bits of info you expect to be there along with other CL arguments?

jayhf:
Ah, that's great to know. I didn't realize. Where exactly can those context be found? I haven't seen any of those fields in the crash logs I'm analyzing. Are there any docs you'd recommend for what information is available where?

tustanivsky:
The command line arguments can be found at captured crash event's page in `Contexts -> Crash Info` section in case you're using Sentry's plugin integration for error handling:

image

If you're sending crashes to Sentry with the default [Crash Reporter](https://docs.sentry.io/platforms/unreal/configuration/setup-crashreporter/) instead then CL args should be listed in `Contexts -> unreal -> custom`.

Also, you can consider enriching the events with [custom contexts](https://docs.sentry.io/platforms/unreal/enriching-events/context/) via the plugin's API if the default one misses some bits of info you're interested in.

jayhf:
Sorry for the slow response here. I think I've figured out why I wasn't seeing the information before: Most of the issues I'm encountering are assertion failures, which ultimately result in the process exiting, but aren't treated as crashes by the Sentry plugin. Interestingly, even the crash context records `IsAssert`, so this seems to be due to a bug or limitation of Sentry's plugin. Perhaps it's even a Linux specific issue. These two issues might be related: https://github.com/getsentry/sentry-unreal/issues/381 and https://github.com/getsentry/sentry-unreal/issues/496, but neither seems like exactly the issue I'm hitting. At any rate, this clearly isn't the right approach so I'll close the PR and I may follow up with an issue later.

jayhf:
Makes sense. I'll keep an eye on that issue, thanks!

PR #700, opened 2024-11-26
Update engine's version-specific code handling

### LINKED ISSUE #676: Improve engine version-specific code handling
Since Sentry plugin supports multiple Unreal Engine versions there is often a need to maintain its version-specific code. For instance, API that allows disabling Unreal's default crash-capturing mechanism appeared only in UE 5.2.

To handle version-specific code we're using `#if`-s combined with `ENGINE_MAJOR_VERSION`/`ENGINE_MINOR_VERSION` macro defined [here](https://github.com/EpicGames/UnrealEngine/blob/40eea367040d50aadd9f030ed5909fc890c159c2/Engine/Source/Runtime/Launch/Resources/Version.h). The existing checks like `#if ENGINE_MAJOR_VERSION == 5` can be improved by making them more future-proof (i.e. for the situation when UE6 arrives) with something like `ENGINE_MAJOR_VERSION != 4`.

Another good example of how to check engine version can be found in [EngineVersionComparison.h](https://github.com/EpicGames/UnrealEngine/blob/40eea367040d50aadd9f030ed5909fc890c159c2/Engine/Source/Runtime/Core/Public/Misc/EngineVersionComparison.h).

tustanivsky:
Closes #676 

#skip-changelog

### DIFF
--- a/plugin-dev/Source/Sentry/Private/Desktop/CrashReporter/SentryCrashContext.cpp
+++ b/plugin-dev/Source/Sentry/Private/Desktop/CrashReporter/SentryCrashContext.cpp
@@ -10,11 +10,11 @@
 #if USE_SENTRY_NATIVE
 
 FSentryCrashContext::FSentryCrashContext(TSharedPtr<FSharedCrashContext> Context)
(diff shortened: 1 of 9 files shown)


PlasmaDev5 (approved):
changes approved

PR #706, opened 2024-12-06
Build sentry from source

PlasmaDev5:
WIP changes to build sentry-native from source.

Currently i have only tested/developed on windows and will not yet fully compile due to not linking libraries.
The changes here are based on UE4CMake with modification and cleanup to better function for the needs of sentry.
going forward further modification will be needed to add platform support for all our needs. 
In addition we will need to load custom toolchains for some platforms.

TODO:
- Link generated libraries
- Android
- IOS
- Linux
- Mac
- Other Platforms


PlasmaDev5:
@tustanivsky Any insights on the build issues. from what i can tell something is referencing `modules/sentry-native` but i did checked everything and did a string search and nothing comes up. It should all have been moved to `plugin-dev/sentry-native`

PlasmaDev5:
Only thing left before final review is add the ability to build for both crashpad and breakpad independently

PlasmaDev5:
I have marked this as ready to review as all contents are ready. The is 1 thing missing related to building backtrace from CI and this looks to be a larger discussion

bruno-garcia:
image
@PlasmaDev5 quite a few CI steps are broken though

PlasmaDev5:
Ah yeah when I looked at it before it looked to be a permissions issue but looks to be doesn't with that last build.  I'll take a look

scripts/packaging/package-marketplace.snapshot
```diff
@@ -8,6 +8,3367 @@ Resources/
 Resources/Icon128.png
 Scripts/
 sentry-cli.properties
+sentry-native/
+sentry-native/.clang-format
+sentry-native/.codechecker-ignore
+sentry-native/.codecov.yml
+sentry-native/.craft.yml
+sentry-native/.editorconfig
+sentry-native/.git
```
bruno-garcia, line 17: Maybe a @tustanivsky question:

Should we be adding `.git` to the packages? This will include all commit history etc.
Similarly about `.github`, looks like we're adding the whole git submodule as-is.

plugin-dev/sentry-native
bruno-garcia, line 1: Did this require a change in `sentry-native` to work?
If so, lets get that added to `main` through a PR on `sentry-native` first. Then we bump `main` here and pull that into this branch. 

This way we don't need to change `sentry-native` in this PR and having it pass CI checks that the change is compatible with it.

plugin-dev/Source/Sentry/Sentry.Build.cs
```diff
@@ -1,16 +1,422 @@
...
+
+    private bool IsUnixPlatform(UnrealTargetPlatform platform) {
+        return platform == UnrealTargetPlatform.Linux || platform == UnrealTargetPlatform.Mac;
+    }
+}
+
 public class Sentry : ModuleRules
 {
+	[CommandLine("-usebreakpad")]
+	public bool bUseBreakpad = false;
```
bruno-garcia, line 418: If I say `false`, what am I using then?
And would that be true regardless of platform? (On Xbox today we only support breakpad so that's odd)

Seems like config will be more complicated than a `bool` as it'll depend on what platform you're targeting, and what's supported there.

plugin-dev/Source/Sentry/Sentry.Build.cs
```diff
@@ -1,16 +1,422 @@
...
+			}
+		}
+		return true;
+	}
+	
+	private string GetWindowsGeneratorName(WindowsCompiler compiler)
+	{
+		string generatorName="";
+
+		switch(compiler)
```
bruno-garcia, line 194: Do we cover all these branches in CI?

I don't think we're building on Windows yet, so we need to add that, for example.

plugin-dev/Source/Sentry/Sentry.Build.cs
```diff
@@ -1,16 +1,422 @@
...
+{
+	public static bool EqualsUpToSeconds(this DateTime dt1, DateTime dt2)
+	{
+		return dt1.Year == dt2.Year && dt1.Month == dt2.Month && dt1.Day == dt2.Day &&
+		       dt1.Hour == dt2.Hour && dt1.Minute == dt2.Minute && dt1.Second == dt2.Second;
+	}   
+}
+public class CMakeTargetInst
+{
+	// Based on UE4CMake with modifications
```
bruno-garcia, line 29: Is there a link to where this came from that you can copy and paste here?

Instead of linking to the file through a branch, use the commit sha (on GitHub, if you press `y` it changes the path from branch to sha automatically.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Features
+
+- Build sentry from source ([#706](https://github.com/getsentry/sentry-unreal/pull/706))
```
bruno-garcia, line 7: Is this accurate?

```suggestion
- Build the plugin through the Unreal Engine build system ([#706](https://github.com/getsentry/sentry-unreal/pull/706))
```

.gitmodules
bruno-garcia, line 1: Did it need to move to `plugin-dev`? It was located on `modules` before and it's unclear why it had to change.

plugin-dev/Source/Sentry/Sentry.Build.cs
PlasmaDev5: I should probably rename this. Its an override more then anything. Native already select the backend needed for selected platforms but this allows the build system to override to build to use breakpad on windows. its needed for shipping releases as we only ship breakpad for the marketplace version

plugin-dev/Source/Sentry/Sentry.Build.cs
PlasmaDev5: We do not currently build windows. i did talk to Ivan about this briefly when gathering information. Last info i got on the topic was during a weekly meeting it was mentioned he was pointed at some stuff that may help get it working.

As for the generators we currently dont build all the unreal generators even on linux. Im not even sure we try all cmake generators on the native side.  But the general goal here is to map the unreal compiler to its best fit in the cmake space.

.gitmodules
PlasmaDev5: The SDK plugin now needs to know the path to native in order to build. I did originally try do this via symlink but with the SDK itself being used as a symlink it caused errors and pointed to the wrong direction. The safest and least error prone option for users was to simple move native into the SDK folder so that wherever you build SDK it can easily locate native without manual user steps

PlasmaDev5:
Paths tested and working on windows. 
Output paths also match when triggering a cross compile with linux.

```Calling: cmd.exe /c cmake.exe -G "Unix Makefiles" -S "D:\Sentry\Unreal\SDK\sentry-unreal\sample/Plugins/Sentry/sentry-native" -B "D:\Sentry\Unreal\SDK\sentry-unreal\sample\Intermediate\CMakeTarget\sentry-native\Linux\build" -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX="D:\Sentry\Unreal\SDK\sentry-unreal\sample\Intermediate\CMakeTarget" -DSENTRY_BUILD_SHARED_LIBS=ON ```

plugin-dev/Source/Sentry/Sentry.Build.cs
bruno-garcia: Yeah then it needs a better name, and docs on it (XML docs over C# members)

plugin-dev/Source/Sentry/Sentry.Build.cs
bruno-garcia: We might be better off adding all of the CI infrastructure first, before we continue with this change.
That way we can be confident we started off with something that works, and the changes continue to do so

.gitmodules
bruno-garcia: It seems that one side effect is that we have the whole repo there (sentry-native) and a bunch of unrelated stuff shows up in the final package. I left notes on those snapshot files.

Also, why is it named `plugin-dev` then? is it relates to dev vs prod time?
Could `sentry-native` be a better name if that's what it is?

bruno-garcia: If that's based on Unity: https://github.com/getsentry/sentry-unity/tree/cb71faad6304e437b7e6d063f198d99c4aa57069/package-dev

We named it like that because we have the package used in development and the one used for packaging and release.

Doesn't seem to be the same situation here

PlasmaDev5:
As discussed i have reverted linux to make this initial version window and xbox only.

PlasmaDev5:
The 5.0 branch is currently failing due to a bug in unreal https://issues.unrealengine.com/issue/UE-143579
I am currently looking into work arounds

tustanivsky (approved):
After updating the process of building the Native SDK in Unreal (#771) we're good to proceed with the merge. Build scripts like `Sentry.Build.cs` and `SentryNative.Build.cs` could benefit from some cleanup but we can tackle that in a separate ticket. Also, we'll need to document how to build the Native SDK from within the Unreal plugin.

PR #726, opened 2025-01-07
Allow upload script to support different bash locations

jayhf:
Using `/usr/bin/env` avoids making assumptions about where the bash executable is installed. See [this stack overflow post](https://stackoverflow.com/questions/21612980/why-is-usr-bin-env-bash-superior-to-bin-bash) for reference


tustanivsky (approved):
@jayhf Thank you!

PR #730, opened 2025-01-10
Add workflow to build docker Win image

bruno-garcia:
triggering CI via a repo to get new snapshot of vars


bruno-garcia:
Passed through that step so the account stuff is resolved

Issue #728, opened 2025-01-10
ios log missing


NoCodeBugsFree:
### Environment - IOS

How do you use Sentry?
Sentry SaaS (sentry.io) or self-hosted/on-premise (which version?) ( both same result )

Which version of the SDK? [0.21.1](https://github.com/getsentry/sentry-unreal/releases/tag/0.21.1)

Which version of Unreal? 5.3

### Steps to Reproduce
project , configured with attach game logs
https://i.imgur.com/vv2dOsL.png
this one forks fine for me on windows, linux, android , but ios crash report have no log attached. 

### Expected Result

ios crash have a log like other platforms did

### Actual Result

no logs for ios crash https://i.imgur.com/UMXmmxd.png

tustanivsky:
@NoCodeBugsFree On iOS, game logs can only be attached to captured events in development (non-shipping) builds. Could you let us know if you have this issue in that configuration as well?

Also, there is another related bug with iOS game log attachments (specific to crash events) - #302

NoCodeBugsFree:
1) I have engine modified to have logs both development/shipping builds
2) even in development builds we dont have logs attached 
3) 100 percent logs exist on devices and I can fetch it both development/shipping builds

tustanivsky:
Did you make any changes to the engine related to log file naming/location?

Is the missing log issue occurring only for crashes, or does it also affect events that are manually captured?

The underlying Sentry Cocoa SDK should always be able to pick up a log file from its default location if the `Attach game log to captured events` option was enabled in project settings:
https://github.com/getsentry/sentry-unreal/blob/d7d7973d645de4f66328d93aff673d4662be33ef/plugin-dev/Source/Sentry/Private/Apple/SentrySubsystemApple.cpp#L57

NoCodeBugsFree:
> or does it also affect events that are manually captured?

tested.
no logs for ios. android is fine

bruno-garcia:
Affecting Supercell too. Can we bump this?

tustanivsky:
@NoCodeBugsFree Do you have `Support iTunes File Sharing` option enabled in your projects `Platform -> iOS` settings? Also, do you override the default log file name via `LOG`/`ABSLOG` command line params or by defining `UE_CUSTOM_LOG_FILENAME`?

I couldn't reproduce this issue so far as log files are always saved under `NSDocumentDirectory` on my iOS device rather than `NSLibraryDirectory`.

tustanivsky:
[...]

PR #732, opened 2025-01-13
temp fix for ios log attach

NoCodeBugsFree:


bitsandfoxes (commented):
~~I'm not sure what the underlying issue is this PR is addressing. Is the SDK not attaching logfiles? Did the logfiles move? Are we attaching the wrong ones? Could you clarify?~~
I missed the linked issue. My bad.

bruno-garcia:
@tustanivsky is this something we need to merge?

tustanivsky:
> @tustanivsky is this something we need to merge?

I've requested additional info in the linked ticket (#728). The changes suggested in this PR feel more like a workaround while the root cause of the issue remains unclear. Also, I wasn't able to reproduce it on my end so far which could mean it's something project/engine-specifc.

EvoPulseGaming:
In UE 5.5 what I found was that IOSPlatformFile checks for DoesCreatePublicFiles() when getting the path for the log.

In FIOSOutputDeviceFile::Serialize where the log is written to disk, it temporarily changes DoesCreatePublicFiles() to true via (&PlatformFile)->SetCreatePublicFiles(true).
It writes the log and then changes it back to whatever the value was before.

The problem with this, is there is no way for any other code to know about this. So when getting the file path to the log, you'll get the wrong path.

tustanivsky:
Superseded by #873

PR #735, opened 2025-01-13
Replace 'convertor' with 'converter'

### LINKED ISSUE #734: "Convertors" seems to be a misspelling of "Converters"
This tripped me up while working with the code because I was attempting to include the wrong files when typing out the word "converter".

Based on [Google Ngram viewer](https://books.google.com/ngrams/graph?content=converter%2Cconvertor&year_start=1800&year_end=2022&corpus=en&smoothing=3&case_insensitive=false) and Oxford English Dictionary, it seems like "converter" is the correct term.

nwhite-riot:
Fixes #734


tustanivsky (approved):
@nwhite-riot Thank you for fixing this!

Issue #733, opened 2025-01-13
Break up `Desktop` and `Apple` submodule into specific platforms


bruno-garcia:
Today our support for Windows are Linux lives on [Desktop](https://github.com/getsentry/sentry-unreal/tree/main/plugin-dev/Source/Sentry/Private/Desktop). And iOS and macOS on [Apple](https://github.com/getsentry/sentry-unreal/tree/main/plugin-dev/Source/Sentry/Private/Apple)

That was done this way since we mainly point to an underlying Sentry SDK (`sentry-native` and `sentry-cocoa` respectively).
But Unreal itself has platform specific subsystems like macOS, Linux, iOS, etc. 

There are several advantages with breaking up the code in the same structure as compared with `#ifdef` throughout the files.

tustanivsky:
Closing this since #745 which improves platform-specific code separation was already merged into `main`.

PR #744, opened 2025-01-21
Remove assertion/ensure callstack modification logic

tustanivsky:
This PR removes custom SDK logic for cutting off redundant callstack frames that prevented issues from being grouped properly at Sentry.

It also restores the default assertion handling flow on desktop which addresses the issue with missing minidump attachment for this type of event.

Closes #669 
Closes #503 

This should be merged after https://github.com/getsentry/sentry/pull/83813

PR #745, opened 2025-01-21
Refactor code to better align with Unreal's structure

nwhite-riot:
Addresses #733 and #754.

The intent of this change is to better align the structure/layout of code with Unreal by separating out platform-specific code in a more meaningful way.

* Split `SentrySubsystemDesktop` into `GenericPlatform`, `Microsoft`, `Windows`, and `Linux`
    * Better encapsulation for platform-specific functionality
    * Prepares for adding Xbox support (since it'll inherit from `Microsoft` and not `Windows`)
* Use `typedef` overrides for multiplatform implementations (e.g. Sentry subsystem)
* Add `HAL` to house platform-specific includes, overrides, and macros
* Where possible, retain existing functionality
    * Unfortunately, even though a number of changes where file moves and renames, Git doesn't seem to recognize and instead shows deletes and create
* Standardized Unreal object creation and native object retrieval into templatized base (with `TSentryImplWrapper<>`)


nwhite-riot:
One thing that worries me a little here is test coverage - it appears test coverage wasn't 100% before the refactor, and now it'll be even lower. @tustanivsky, what's the preferred approach for adding/managing tests?

plugin-dev/Source/Sentry/Private/Tests/SentryEvent.spec.cpp
```diff
@@ -15,7 +17,8 @@ void SentryEventSpec::Define()
 {
 	BeforeEach([this]()
 	{
-		SentryEvent = NewObject<USentryEvent>();
+		TSharedPtr<ISentryEvent> NativeImpl = MakeShareable(new FGenericPlatformSentryEvent);
+		SentryEvent = USentryEvent::Create(NativeImpl);
```
j0tt, line 21: Do we need to capture the shared ptr?
```suggestion
		SentryEvent = USentryEvent::Create(MakeShareable(new FGenericPlatformSentryEvent));
```

plugin-dev/Source/Sentry/Public/SentryBreadcrumb.h
```diff
@@ -58,10 +57,4 @@ class SENTRY_API USentryBreadcrumb : public UObject
 	/** Gets the level of the breadcrumb. */
 	UFUNCTION(BlueprintPure, Category = "Sentry")
 	ESentryLevel GetLevel() const;
-
-	void InitWithNativeImpl(TSharedPtr<ISentryBreadcrumb> breadcrumbImpl);
-	TSharedPtr<ISentryBreadcrumb> GetNativeImpl();
-
-private:
-	TSharedPtr<ISentryBreadcrumb> BreadcrumbNativeImpl;
 };
```
j0tt, line 60: /r/n

plugin-dev/Source/Sentry/Public/SentryUserFeedback.h
```diff
@@ -44,10 +41,4 @@ class SENTRY_API USentryUserFeedback : public UObject
 	/** Gets comments of the user about what happened. */
 	UFUNCTION(BlueprintPure, Category = "Sentry")
 	FString GetComment() const;
-
-	void InitWithNativeImpl(TSharedPtr<ISentryUserFeedback> userFeedbackImpl);
-	TSharedPtr<ISentryUserFeedback> GetNativeImpl();
-
-private:
-	TSharedPtr<ISentryUserFeedback> UserFeedbackNativeImpl;
 };
```
j0tt, line 44: /r/n

j0tt (commented):
I still want to better understand the macro magic here, but I'll leave that for tomorrow. This is a first pass review :) I don't feel strongly about ending a file with a newline if you're not keen on the busy work, it's just a habit of mine 😅

plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentrySubsystem.cpp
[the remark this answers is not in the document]
nwhite-riot: Unfortunately there are a large number of changes :(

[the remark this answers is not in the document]
j0tt: Okay, I'll read the whole file 😆 😭

plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentrySubsystem.h
```diff
@@ -52,18 +54,38 @@ class SentrySubsystemDesktop : public ISentrySubsystem
...
-	TSharedPtr<SentryScopeDesktop> GetCurrentScope();
+	TSharedPtr<FGenericPlatformSentryScope> GetCurrentScope();
+
+protected:
+	virtual void ConfigureTransport(sentry_options_t* Options) {}
+	virtual void ConfigureHandlerPath(sentry_options_t* Options) {}
+	virtual void ConfigureDatabasePath(sentry_options_t* Options) {}
+	virtual void ConfigureLogFileAttachment(sentry_options_t* Options) {}
+	virtual void ConfigureScreenshotAttachment(sentry_options_t* Options) {}
+	virtual void ConfigureGpuDumpAttachment(sentry_options_t* Options) {}
```
j0tt, line 65: can you mark these as overrides? I went to the cpp file to see the implementation because I didn't notice they were `{}` empty.
and preferably remove the virtual?

plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentrySubsystem.h
```diff
@@ -52,18 +54,38 @@ class SentrySubsystemDesktop : public ISentrySubsystem
...
 
-private:
 	FString GetHandlerPath() const;
 	FString GetDatabasePath() const;
 	FString GetScreenshotPath() const;
+	virtual int32 GetEnsureFramesToSkip() const { return 7; }
+	virtual int32 GetAssertionFramesToSkip() const { return 5; }
+	virtual FString GetHandlerExecutableName() const { return TEXT("invalid"); }
```
j0tt, line 72: where do these magic numbers come from?

plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentrySubsystem.cpp
```diff
@@ -0,0 +1,569 @@
...
+
+	sentry_options_set_dsn(options, TCHAR_TO_ANSI(*settings->Dsn));
+	sentry_options_set_environment(options, TCHAR_TO_ANSI(*settings->Environment));
+	sentry_options_set_logger(options, PrintVerboseLog, nullptr);
+	sentry_options_set_debug(options, settings->Debug);
+	sentry_options_set_auto_session_tracking(options, settings->EnableAutoSessionTracking);
+	sentry_options_set_sample_rate(options, settings->SampleRate);
+	sentry_options_set_max_breadcrumbs(options, settings->MaxBreadcrumbs);
+	sentry_options_set_before_send(options, HandleBeforeSend, this);
+	sentry_options_set_on_crash(options, HandleOnCrash, this);
```
j0tt, line 238: Looks like this used to be before instead of on (`sentry_options_set_on_crash(options, HandleBeforeCrash, this);`)

Are _we_ wrong or was _it_ wrong? Or something more obscure?

plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentrySubsystem.h
j0tt: I found their origin in SentrySubsystemDesktop.cpp
```
TSharedPtr<ISentryId> SentrySubsystemDesktop::CaptureAssertion(const FString& type, const FString& message)
{
#if PLATFORM_WINDOWS
	int32 framesToSkip = 7;
#else
	int32 framesToSkip = 5;
#endif
```

So... I guess it's no different than what it was before, but maybe a comment? I didn't see the overrides in other platforms but I assume they exist.

plugin-dev/Source/Sentry/Private/Null/NullSentryAttachment.h
```diff
@@ -0,0 +1,14 @@
...
+
+class FNullSentryAttachment : public ISentryAttachment
+{
+public:
+	virtual ~FNullSentryAttachment() = default;
+
+	virtual TArray<uint8> GetData() const { return {}; }
+	virtual FString GetPath() const { return TEXT(""); }
+	virtual FString GetFilename() const { return TEXT(""); }
+	virtual FString GetContentType() const { return TEXT(""); }
```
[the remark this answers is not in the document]
nwhite-riot: The codebase has both `virtual` and `override`, so I've stuck with that for now and updated the code!

plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentrySubsystem.h
nwhite-riot: They're not marked as overrides because they're not overrides - they're the base default implementation (i.e. no-op) because not all of them will be overridable for all derived platforms.

plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentrySubsystem.h
nwhite-riot: What are you looking for in a comment? They're self-documenting by nature of the method names, and the comment would only be relevant for people looking for where they originally came from, not their behavior etc.

plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentrySubsystem.cpp
nwhite-riot: The Sentry function is `...on_crash`, so I updated the function it calls to match (since `...before_send` calls `HandleBeforeSend`)

tustanivsky:
@nwhite-riot Thank you for all the effort you've put into this refactoring so far! Overall, things are looking good and we appreciate making the SDK more future-proof in terms of adding consoles support. This review might require a few iterations given the amount of changes however I'd like to share some of my early thoughts:

One of my major concerns is that Sentry entities (e.g. event, user, breadcrumb) cannot be created directly with `NewObject` in client code as they now have to be instantiated using the `TSentryImplWrapper::Create` static method which might not be as intuitive. This also requires a pointer to the object's native implementation to properly initialize its Unreal wrapper however the necessary interfaces are currently inaccessible outside the plugin. So the only way to create a valid Sentry object at this point is through `USentryLibrary`. 

Previously we relied at constructors and/or initialize methods to ensure that the default native impl is set whenever possible making Unreal objects self-contained. This approach wasn't ideal but it seemed more error-proof for SDK users. Perhaps it would make sense to combine this with the object construction utilities introduced here?

If so, there doesn't seem to be any other significant changes to the plugin's public API and therefore preserving its backward compatibility with older versions would be preferred.

tustanivsky:
> One thing that worries me a little here is test coverage - it appears test coverage wasn't 100% before the refactor, and now it'll be even lower. @tustanivsky, what's the preferred approach for adding/managing tests?

After adding the `CreateSharedSentry___` methods to `HAL` it should be possible to maintain the same level of test coverage as before, right? Our current CI pipeline is limited to running checks on Linux but we're actively working on expanding it to support more targer platforms.

plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentrySubsystem.h
tustanivsky: Just a quick heads-up on this - we're removing the event's callstack modification logic from the SDK as part of https://github.com/getsentry/sentry-unreal/pull/744. This was a temporary workaround to address invalid grouping of assertion events in Sentry

plugin-dev/Source/Sentry/Public/SentrySettings.h
```diff
@@ -279,19 +279,19 @@ class SENTRY_API USentrySettings : public UObject
 	bool UseProxy;
 
 	UPROPERTY(Config, EditAnywhere, BlueprintReadWrite, Category = "General|Transport",
-		Meta = (DisplayName = "HTTP proxy (for Windows/Linux only)", ToolTip = "HTTP proxy through which requests can be tunneled to Sentry.", EditCondition = "UseProxy"))
```
tustanivsky, line 282: Why was this removed? Having an explicit notice about platforms for which this setting is relevant could help to avoid confusion among users.

plugin-dev/Source/Sentry/Public/SentrySettings.h
```diff
@@ -279,19 +279,19 @@ class SENTRY_API USentrySettings : public UObject
...
-	UPROPERTY(Config, EditAnywhere, Category = "General|Desktop",
+	UPROPERTY(Config, EditAnywhere, Category = "General|Windows",
 		Meta = (DisplayName = "Override Windows default crash capturing mechanism (UE 5.2+)", ToolTip = "Flag indicating whether to capture crashes automatically on Windows as an alternative to Crash Reporter."))
 	bool EnableAutoCrashCapturing;
 
-	UPROPERTY(Config, EditAnywhere, Category = "General|Desktop",
-		Meta = (DisplayName = "Sentry database location (for Windows/Linux only)", ToolTip = "Location where Sentry stores its internal/temporary files."))
+	UPROPERTY(Config, EditAnywhere, Category = "General",
+		Meta = (DisplayName = "Sentry database location", ToolTip = "Location where Sentry stores its internal/temporary files."))
 	ESentryDatabaseLocation DatabaseLocation;
```
tustanivsky, line 295: This setting makes sense only on Windows/Linux where we use `sentry-native`. Why moving it to "General" section?

nwhite-riot:
> So the only way to create a valid Sentry object at this point is through `USentryLibrary`
I thought this would be preferable to having users instantiate their own objects - either we have the library to do it for them, or we remove the library and allow them to construct the objects manually. Feels odd to have both?

plugin-dev/Source/Sentry/Public/SentrySettings.h
nwhite-riot: Originally it was `General|Desktop`, so I wasn't sure how we wanted to handle it now that `Desktop` doesn't exist. Perhaps it becomes `General|Native` or something similar?

plugin-dev/Source/Sentry/Public/SentrySettings.h
nwhite-riot: If it's Windows and Linux only, should it not have been in `General|Desktop` like the others? I removed it because it wasn't clear about limitations. Perhaps we need to split the settings up to make it more obvious which ones are exclusive to which platforms?

plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentrySubsystem.h
j0tt: Ahh, I wrongly assumed they were coming from ISentrySubsystem.

plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentrySubsystem.cpp
j0tt: I see now we renamed the method for consistency in naming, we good here.

tustanivsky:
> I thought this would be preferable to having users instantiate their own objects - either we have the library to do it for them, or we remove the library and allow them to construct the objects manually. Feels odd to have both?

`USentryLibrary` was originally designed as a convenience utility for users working with the plugin's API in Blueprints. At the same time, manually constructing objects with `NewObject `is a common approach that provides additional flexibility for creating various Sentry instances in C++. I believe keeping both options available will help developers avoid potential issues when switching to this new SDK version.

nwhite-riot:
> I believe keeping both options available will help developers avoid potential issues when switching to this new SDK version.

My last question about this is around object validity: if a user calls `NewObject<...>()` but then does not initialize with the native implementation, the object would be invalid and all methods would no-op (without any kind of logging). It feels to me like this is an anti-pattern that will end up causing more problems.

I will look into retaining the desired functionality (of allowing `NewObject<...>()`).

Thanks for the discussion!

tustanivsky:
> if a user calls NewObject<...>() but then does not initialize with the native implementation, the object would be invalid and all methods would no-op (without any kind of logging). It feels to me like this is an anti-pattern that will end up causing more problems.

For most Sentry classes that wasn't an issue since their default constructor handled the native implementation initialization (except `USentryAttachment` and maybe few others where that's not possible). Will it make sense to add extra logging during the `NativeImpl` validation?

plugin-dev/Source/Sentry/Private/Tests/SentryBreadcrumb.spec.cpp
```diff
@@ -15,7 +17,7 @@ void SentryBreadcrumbSpec::Define()
 {
 	BeforeEach([this]()
 	{
-		SentryBreadcrumb = NewObject<USentryBreadcrumb>();
+		SentryBreadcrumb = USentryBreadcrumb::Create(MakeShareable(new FGenericPlatformSentryBreadcrumb));
```
tustanivsky, line 20: Can we use `CreateSharedSentryBreadcrumb()` from `HAL/PlatformSentryBreadcrumb.h` here so that this test (and others too) won't be limited to Win/Linux platforms?

nwhite-riot: This should now be resolved!

plugin-dev/Source/Sentry/Private/SentryBeforeSendHandler.cpp
```diff
@@ -16,21 +17,21 @@ USentryEvent* USentryBeforeSendHandler::HandleBeforeSend_Implementation(USentryE
...
 	const USentrySettings* Settings = FSentryModule::Get().GetSettings();
 
-	if(Settings->EnableAutoLogAttachment)
+	if (Settings->EnableAutoLogAttachment && Hint != nullptr)
 	{
 #if PLATFORM_ANDROID || PLATFORM_APPLE
 		const FString LogFilePath = Event->IsCrash() ? SentryFileUtils::GetGameLogBackupPath() : SentryFileUtils::GetGameLogPath();
 #else
 		const FString LogFilePath = SentryFileUtils::GetGameLogPath();
 #endif
```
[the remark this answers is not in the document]
nwhite-riot: I don't think I have the brainpower to find a good way of separating this, so I'd like to leave it for now and reconsider it in the future

plugin-dev/Source/Sentry/Public/SentrySettings.h
nwhite-riot: I moved it under `Native` since it's a better representation of what's actually affected

tustanivsky:
I've created a PR (https://github.com/nwhite-riot/sentry-unreal/pull/3) that addresses some of the build issues on Apple/Android platforms.

tustanivsky (approved):
Thank you @nwhite-riot!

PR #751, opened 2025-01-22
Change how variables are piped to fix ID retrieval

### LINKED ISSUE #750: `findCiRun` in `download-sdks.ps1` fails to find correct values
The expected output of `findCiRun` is to discover a successful `package-plugin-workflow` workflow completion and download the corresponding binaries. Unfortunately, the code as written outputs nothing due to the way the variables are piped together.

I'm running this as part of `init-win.ps1`.

Here's the output of my `$PSVersionTable`:

```powershell
Name                           Value
----                           -----
PSVersion                      5.1.22621.4391
PSEdition                      Desktop
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
BuildVersion                   10.0.22621.4391
CLRVersion                     4.0.30319.42000
WSManStackVersion              3.0
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
```

nwhite-riot:
Fixes #750.

### DIFF
--- a/scripts/download-sdks.ps1
+++ b/scripts/download-sdks.ps1
@@ -5,8 +5,8 @@ Write-Host "Downloading native SDKs from the latest CI pipeline"
 function findCiRun([string] $branch)
 {
     Write-Host "Looking for the latest successful CI run on branch '$branch'"
-    $id = gh run list --branch $branch --workflow package-plugin-workflow --json 'conclusion,databaseId' `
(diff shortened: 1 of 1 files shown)


tustanivsky (approved):
That's a great catch @nwhite-riot! It turns out this script was working as expected only with PowerShell 7. With this change users can now run it with PowerShell 5 which is the default version. Thank you for making it more accessible!

PR #752, opened 2025-01-23
Add missing changelog entry

tustanivsky:
This PR adds a missing changelog entry for #751

#skip-changelog


tustanivsky:
This change seems to be irrelevant for plugin end users so there is no need to add it to changelog.

PR #756, opened 2025-01-29
Avoid trying to use Slate for screenshots in Sentry during a crash if not on the game thread or slate thread, to avoid causing errors in Slate.

### LINKED ISSUE #755: Assertion during capturing a screenshot while handling a crash.
If Sentry tries to call CaptureScreenshot() while handling a crash from a thread that can't use Slate, it will hit an assert in SWidget.cpp that checks for IsInGameThread() || IsInSlateThread().

A fix is to check for this situation prior to calling Slate's TakeScreenshot().

Tested on UE5.4 Win64.

swbrownuwe:
Check if we're on a thread that can use Slate to capture a screenshot before attempting to do so.

Fixes #755


tustanivsky:
That's a good catch! Thank you @swbrownuwe

Issue #757, opened 2025-01-29
Ability to get the event id for use in local logs


swbrownuwe:
Would like to be able to get the event id so that a URL to what the event will be on Sentry's site can be left in the game's log file when handling an error.

Didn't notice a way to do this as-is, but it's easy enough for Desktop at least to add something that calls sentry_value_get_by_key() for event_id.

bitsandfoxes:
Thanks for the feature request! I think this is a great idea!
Looking at this from the perspective of the Unreal SDK being a hybrid SDK we could investigate something like a `GetLastId` API to retrieve the ID.

PR #758, opened 2025-01-29
Add a way to get the event ID of an event for Desktop.

swbrownuwe:
Partial implementation of adding a GetId() to ISentryEvent, implemented only for Desktop here to return the event_id from sentry-native.  Haven't looked into how this would be implemented on Android, Mac, etc..  Provided as a working example for Desktop only, not to merge as-is.

Addresses #757

### DIFF
--- a/plugin-dev/Source/Sentry/Private/Desktop/SentryEventDesktop.cpp
+++ b/plugin-dev/Source/Sentry/Private/Desktop/SentryEventDesktop.cpp
@@ -2,6 +2,8 @@
 
 #include "SentryEventDesktop.h"
 #include "SentryDefines.h"
+#include "SentryId.h"
+#include "SentryIdDesktop.h"
 
 #include "Infrastructure/SentryConvertersDesktop.h"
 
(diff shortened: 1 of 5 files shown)


tustanivsky:
@swbrownuwe Thank you for this PR - adding `GetId()` for the event looks reasonable. Unfortunately, we can't accept it as-is since the suggested changes will break the build for other platforms (Apple, Android). Would you be open to adding platform-specific implementations for them as well? I believe checking the following APIs is a good place to start:
- Android: https://github.com/getsentry/sentry-java/blob/2fc1ed7633b3741bc3d8a49b65888cdb95ea5300/sentry/src/main/java/io/sentry/SentryBaseEvent.java#L128
- Cocoa: https://github.com/getsentry/sentry-cocoa/blob/7c8df469903b729a05605cf9f993670767e69720/Sources/Sentry/Public/SentryEvent.h#L30

The Unreal SDK has plenty of examples of how to query data from the native layers which could be helpful for this task.

tustanivsky:
Superseded by #768

PR #776, opened 2025-02-07
Remove obsolete `enableTracing` option usage

### LINKED ISSUE #765: Remove `enableTracing` setting
Even though tracing itself is handled via the underlying Native, Cocoa and Android SDKs we need to remove the corresponding setting from Unreal plugin as well.

See https://github.com/getsentry/team-sdks/issues/84 for more details.

tustanivsky:
The only place where the obsolete `enableTracing` option was still used for underlying native SDK initialization was the Android wrapper for `sentry-java`.

On the Unreal side, this flag is present as it's needed to render a corresponding checkbox in the plugin's settings menu, which is the default tool for configuring Sentry.

image

Closes #765

#skip-changelog


bitsandfoxes (approved):

bitsandfoxes:
Just for documentation's sake: Even tho this is a breaking change, because the UE SDK is still in a pre-1.0 state I'm fine with removing this without a deprecation-period.

tustanivsky:
> Just for documentation's sake: Even tho this is a breaking change, because the UE SDK is still in a pre-1.0 state I'm fine with removing this without a deprecation-period.

@bitsandfoxes This PR doesn't break anything on Unreal's side of things as we're not removing `EnableTracing` flag from the plugin settings menu. The change made is entirely related to the UE plugin’s internals.

PR #781, opened 2025-02-11
Split build platform

PlasmaDev5:
Each platform now has its own module


tustanivsky (changes requested):
It looks a bit strange how git included those new platform build scrips - they all should belong to `plugin-dev` rather than `sample`. Technically, these files are not part of the plugin at the moment so the package content validation check passes successfully even though it's incomplete.

PlasmaDev5:
good catch. no idea why git added them that way. ill look into it

tustanivsky (approved):

PR #782, opened 2025-02-12
Fix stack overflow when calling `beforeSend` during object post-loading on mobile

tustanivsky:
This PR applies the fix introduced in #589 to mobile platforms.


plugin-dev/Source/Sentry/Private/Android/Jni/SentryJniAndroid.cpp
```diff
@@ -38,7 +39,14 @@ JNI_METHOD jobject Java_io_sentry_unreal_SentryBridgeJava_onBeforeSend(JNIEnv* e
 	USentryEvent* EventToProcess = USentryEvent::Create(MakeShareable(new SentryEventAndroid(event)));
 	USentryHint* HintToProcess = USentryHint::Create(MakeShareable(new SentryHintAndroid(hint)));
 
-	return handler->HandleBeforeSend(EventToProcess, HintToProcess) ? event : nullptr;
+	USentryEvent* ProcessedEvent = EventToProcess;
+	if (!FUObjectThreadContext::Get().IsRoutingPostLoad)
```
bitsandfoxes, line 43: Can't we skip the whole method in that case?

bitsandfoxes: Nitpick but I'd probably invert the check, bail early and print some debug logs. So users have a way to follow why their beforeSend is not getting called.

tustanivsky: Fair enough, the check is now inverted and prints a log message explaining why custom `beforeSend` handling was skipped.

bitsandfoxes (approved):
Thanks!

PR #791, opened 2025-02-20
Clean up Native SDK build script

### LINKED ISSUE #772: clean up in SentryNative.Build.cs
This issue stems out of discussion with Ivan. 
We would benefit from doing a cleanup/maintenance pass on SentryNative.Build.cs.

tustanivsky:
This PR refactors the script used to build the Native SDK in Unreal improving its clarity. It also removes unnecessary `sentry-native` files (e.g., tests, examples) from the plugin package.

Closes #772

#skip-changelog

### DIFF
--- a/plugin-dev/Source/Sentry/Sentry.Build.cs
+++ b/plugin-dev/Source/Sentry/Sentry.Build.cs
@@ -95,7 +95,7 @@ public class Sentry : ModuleRules
 
 			if (bBuildNativeForWindows)
 			{
-				PrivateDependencyModuleNames.Add("SentryWin64Native");
+				PrivateDependencyModuleNames.Add("SentryNative");
 			}
 			else
 			{
--- a/plugin-dev/Source/ThirdParty/Native/SentryNative.Build.cs
+++ b/plugin-dev/Source/ThirdParty/Native/SentryNative.Build.cs
@@ -23,13 +23,13 @@ public class CMakeTargetInst
 	//
 	// Copyright (c) 2020 Krazer
 	//
-	// 	Permission is hereby granted, free of charge, to any person obtaining a copy
-	// 	of this software and associated documentation files (the "Software"), to deal
+	// Permission is hereby granted, free of charge, to any person obtaining a copy
+	// of this software and associated documentation files (the "Software"), to deal
 	// in the Software without restriction, including without limitation the rights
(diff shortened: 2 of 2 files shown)


plugin-dev/Source/Sentry/Sentry.Build.cs
```diff
@@ -95,7 +95,7 @@ public Sentry(ReadOnlyTargetRules Target) : base(Target)
 
 			if (bBuildNativeForWindows)
 			{
-				PrivateDependencyModuleNames.Add("SentryWin64Native");
+				PrivateDependencyModuleNames.Add("SentryNative");
```
nwhite-riot, line 98: This is likely to conflict with the `SentryNative` module already in the Unreal Engine

plugin-dev/Source/ThirdParty/Native/SentryNative.Build.cs
```diff
@@ -320,215 +324,130 @@ private string CreateCMakeConfigCommand(ReadOnlyTargetRules target, ModuleRules
...
-						"crashpad_zlib.lib"));
-					PublicAdditionalLibraries.Add(Path.Combine(crashpadBuildPath, "tools", "Release",
-						"crashpad_tools.lib"));
-					PublicAdditionalLibraries.Add(Path.Combine(crashpadBuildPath, "util", "Release",
-						"crashpad_util.lib"));
-				}
-			}
+			string installPath = Path.Combine(projectIntermediatePath, "CMakeTarget", "sentry-native", Target.Platform.ToString(), "install");
+
+			RuntimeDependencies.Add(Path.Combine(platformBinariesPath, "crashpad_handler.exe"), Path.Combine(installPath, "bin", "crashpad_handler.exe"));
```
nwhite-riot, line 438: If you're limiting this to Crashpad, do you need to configure the transport in the CMake arguments? What if the user wants Breakpad?

tustanivsky: You can configure which backend and transport to use independently. By default, CMake relies on `winhttp` transport for Windows and there's no need to set it explicitly via the arguments list.

Currently, only `crashpad` is supported when building `sentry-native` from source. However, we can iterate on this and add `breakpad` as a next step.

nwhite-riot (approved):

Issue #722, opened 2025-01-02
Add Windows target support to CI pipeline


tustanivsky:
Windows is a primary target platform for the Sentry Unreal Engine SDK and as part of the ongoing efforts to improve the existing CI pipeline (tracked in [#575](https://github.com/getsentry/sentry-unreal/issues/575)) we need to ensure that it’s fully supported. This will also help maintain reliability and streamline development workflows for Windows-specific features and bug fixes.

There’s no straightforward way to setup Unreal Engine on the standard GitHub runner as the Epic Games Launcher cannot be used for this purpose within the CI environment. Therefore, we need to either create Docker images with Unreal Engine pre-installed or identify a way to access pre-built engine binaries for manual installation.

Outlined below are several approaches to consider (each with its own benefits and limitations) that could help achieve this.

### Official docker images

Epic Games provides [official Docker images](https://github.com/orgs/epicgames/packages/container/package/unreal-engine) for UE that can be categorized into two types: `runtime` and `development` (see [Overview of Containers in Unreal Engine](https://dev.epicgames.com/documentation/en-us/unreal-engine/overview-of-containers-in-unreal-engine#developmentandruntimecontainerimages)).

**Runtime Images:**

The Windows runtime images include only the dependencies needed to run packaged Unreal Engine projects. While these images are useful for tasks like running smoke tests they are not suitable for building the plugin because they lack the required engine’s development tools and libraries.

**Development Images:**

Development images contain the full toolset required for building and testing Unreal Engine plugins and projects. However, these images are currently available only for Linux. We are already utilizing Linux-based development images in our CI pipeline but the lack of Windows equivalent presents a significant challenge for supporting this target platform and there's no clear timeline when it might be released.

### ue4-docker project

An alternative solution is the [ue4-docker project](https://github.com/adamrehn/ue4-docker) which provides Dockerfiles for building both Linux and Windows images. The accompanying documentation ([ue4-docker Docs](https://ue4-docker.docs.adamrehn.com/)) comprehensively outlines the process of building these images and common pitfalls.

Key considerations for integrating ue4-docker into Sentry’s plugin CI pipeline:
- To comply with the [Unreal Engine EULA](https://www.unrealengine.com/en-US/eula/unreal) any Docker images produced using ue4-docker must remain private. This is important to keep in mind when distributing these images via [Sentry's GitHub Package Registry](https://github.com/orgs/getsentry/packages) or any other platform.
[...]

PR #792, opened 2025-02-21
chore: Build Windows UE Docker images in CI

tustanivsky:
Related to #722 

#skip-changelog


.github/workflows/ci.yml
```diff
@@ -1,6 +1,10 @@
 name: package-plugin-workflow
 
-on: push
+on: 
+  push:
+    paths-ignore:
+      - "**.md"
+      - "**ue-docker.yml"
```
bitsandfoxes, line 7: This is to be ignored? I'd assume that changes there would be the trigger for this workflow.

tustanivsky: There's no need to run CI checks when modifying the `ue-docker.yml` workflow as they are not affected by the result of Docker build. That's because Docker image build takes significantly longer than an average CI run and the new/updated images simply won't be available at that moment when they're consumed by `ci.yml`.

.github/workflows/ue-docker.yml
```diff
@@ -0,0 +1,79 @@
...
+        required: true
+        type: choice
+        options:
+        - 4.27
+        - 5.0
+        - 5.1
+        - 5.2
+        - 5.3
+        - 5.4
+        - 5.5
```
vaind, line 17: Is it worth having options here instead of letting user specify it (which would mean we don't need to change this file when a new version is built?

.github/workflows/ue-docker.yml
```diff
@@ -0,0 +1,79 @@
...
+  build:
+    runs-on: windows-latest-16-cores
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      # Building docker images for UE 5.4 and newer require extra disk space
+      - name: Clean up disk space before Docker build        
+        if: ${{ inputs.ue_version == '5.4' || inputs.ue_version == '5.5' }}
```
vaind, line 29: looks like this is going to be necessary all the time, going forward so maybe worth removing the condition.

vaind (approved):

bruno-garcia (commented):
We need `DOCKER_USERNAME` and `DOCKER_TOKEN` added to the repo?

tustanivsky:
> We need `DOCKER_USERNAME` and `DOCKER_TOKEN` added to the repo?

No, we already have these set up and [use](https://github.com/getsentry/sentry-unreal/blob/19afa0a3127410330d6d6928a5131ae689cb971f/.github/workflows/ci.yml#L185) them in `ci.yml`.

.github/workflows/ue-docker.yml
tustanivsky: Let's keep the options for now as they make it clear which UE versions are supported and ensure that developers patch only the corresponding Unreal fork branches used for building the Docker images.

tustanivsky:
[Test workflow run](https://github.com/getsentry/sentry-unreal/actions/runs/13473005875/job/37648795141) revealed a network-related issue which has to be addressed before we can move on with merging this PR. It might be related to the large GitHub runner configuration changes since a similar workflow passed the prerequisites build stage successfully earlier.

PR #793, opened 2025-02-21
Remove sh from Symbol Upload Call

jayhf:
Fixes a build bug we experienced where sh complained about some bash only syntax. This fixes the issue by removing the explicit sh call so that the shebang can do its job and make bash run the script.


tustanivsky:
Thanks @jayhf, this looks like a reasonable addition to the changes you introduced earlier in #726.

PR #794, opened 2025-02-22
Rearrange upload script to check of automatic upload earlier

nwhite-riot:
The upload symbol scripts look at `DefaultEngine.ini` instead of platform-specific Engine config files, so there is no need to check for the platform (and for the CLI executable) before checking to see if automatic upload is even enabled.

This avoids the situation where a new platform (e.g. Xbox) is not configured in the upload script, which would cause a warning to appear in build logs even if upload is disabled.

### DIFF
--- a/plugin-dev/Scripts/upload-debug-symbols-win.bat
+++ b/plugin-dev/Scripts/upload-debug-symbols-win.bat
@@ -20,18 +20,9 @@ if "%TargetType%"=="Editor" (
     exit /B 0
 )
 
-if "%TargetPlatform%"=="Win64" (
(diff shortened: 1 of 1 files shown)


nwhite-riot:
Realistically, the upload script should not solely rely on `DefaultEngine.ini` and should mirror the behavior of Unreal (where `.ini` files are layered over each other, from `Base` -> `Default` -> platform-specific). @tustanivsky, perhaps this should be an issue that is investigated in the future?

tustanivsky (approved):
Thank you @nwhite-riot - these changes make sense as we prepare to support more platforms.
Additionally, I've created a separate issue to explore how we can utilize platform-specific `.ini` files as you've suggested.

PR #812, opened 2025-03-07
chore: Add `EngineVersion` key to plugin descriptor only for marketplace version

### LINKED ISSUE #811: `EngineVersion` causes warning for licensee builds
When integrating the plugin in licensee versions of Unreal (i.e. versions build locally instead of through Epic launcher etc.), a warning is raised because the `EngineVersion` value in the plugin descriptor - `5.4.0` - conflicts with the patch version for the version we're using (`5.4.4`).

I'm not sure if there's a better solution for this, or if we need it at all.

Specifically, this is an issue for us because we treat warnings as errors.

tustanivsky:
This PR updates the packaging script to add the `EngineVersion` key to the `Sentry.uplugin` descriptor only for the marketplace version of the plugin.

For UE Marketplace publishing only the major engine version (e.g., `5.5.0`) needs to be specified. In licensee versions of Unreal the patch version may differ from zero which can trigger unnecessary warnings that some organizations treat as errors.

Closes #811 

#skip-changelog

### DIFF
--- a/scripts/packaging/pack.ps1
+++ b/scripts/packaging/pack.ps1
@@ -57,6 +57,8 @@ function packFiles([string] $publishingPlatform)
         $packageName = "sentry-unreal-$version-engine$engineVersion-$publishingPlatform.zip"
         Write-Host "Creating a release package for Unreal $engineVersion as $packageName"
 
+        $newPluginSpec = $pluginSpec
+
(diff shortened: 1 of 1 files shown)


bitsandfoxes (approved):

bruno-garcia:
@tustanivsky something else worth noting on the documentation listng the difference between marketplace and github builds

PR #831, opened 2025-03-17
(Early) spring cleaning

nwhite-riot:
* Renamed `FSentryOutputDeviceError` -> `FSentryErrorOutputDevice` to match Epic's naming convention
* Made some methods `const` on `FSentrySubsystem` where appropriate
* Made spacing/braces etc. consistent in `FSentrySubsystem`
* Updated lambdas to be weak to implicitly handle broadcast after destruction (and to better represent expected lifetime)
* Added `check` and `verify` calls for objects/pointers that should be valid (so that it breaks during development)


plugin-dev/Source/Sentry/Private/SentrySubsystem.cpp
```diff
@@ -121,21 +127,12 @@ void USentrySubsystem::Initialize()
...
 	ConfigureOutputDevice();
+	ConfigureErrorOutputDevice();
 
-#if PLATFORM_WINDOWS
-	if (FEngineVersion::Current().GetMajor() == 5 && FEngineVersion::Current().GetMinor() >= 2)
+	OnEnsureDelegate = FCoreDelegates::OnHandleSystemEnsure.AddWeakLambda(this, [this]()
 	{
-		if (Settings->EnableAutoCrashCapturing)
-		{
-			ConfigureOutputDeviceError();
```
nwhite-riot, line 130: @tustanivsky, since the code in `ConfigureOutputDeviceError()` no longer handles anything other than on Android, I removed the `#if` checks - let me know if that doesn't make sense. Ideally we would create platform-specific error output devices to encapsulate this logic, but that's a change for later

tustanivsky: Fair point, we can safely remove those checks now.

PR #836, opened 2025-03-19
Allow Sentry CLI to authenticate via environment variables during debug symbols upload

### LINKED ISSUE #833: Configure Debug Information File upload using environment variables.
When uploading debug informaiton files automatically with the UE SDK, it only looks for information in the sentry.properties file: https://docs.sentry.io/platforms/unreal/configuration/debug-symbols/

When trying to configure it only with environment variables (similar to the sentry-cli) the process returns an error and exits:
```
Warning: Sentry: Properties file is missing
```

From the [upload scripts in here](https://github.com/getsentry/sentry-unreal/tree/main/plugin-dev/Scripts) there is no option to configure it only with environment variables. 

This request to add support to environment variables, or to clearly document it if it is already supported.

tustanivsky:
This PR extends the debug symbols upload settings by allowing Sentry CLI to be configured using either a `sentry.properties` file or environment variables (`SENTRY_AUTH_TOKEN`, `SENTRY_ORG` and `SENTRY_PROJECT`). Users can now select their preferred configuration method in Sentry plugin settings.

Closes #833

Corresponding documentation update: https://github.com/getsentry/sentry-docs/pull/13045

unreal-debug-symbols


plugin-dev/Source/Sentry/Sentry_Android_UPL.xml
```diff
@@ -48,8 +51,16 @@
         <copyFile src="$S(PluginDir)/../ThirdParty/Android/sentry-native-ndk-release.aar" dst="$S(BuildDir)/gradle/app/libs/sentry-native-ndk-release.aar" />
         <if condition="bUploadSymbols">
             <true>
-                <copyFile src="$S(ProjectDir)/sentry.properties" dst="$S(BuildDir)/gradle/sentry.properties" />
-                <copyFile src="$S(ProjectDir)/sentry.properties" dst="$S(BuildDir)/gradle/AFSProject/app/sentry.properties" />
+                <if condition="bUseEnvVariables">
+                    <true>
+                        <log text="Sentry gradle plugin will use environment variables to authenticate with Sentry"/>
```
nwhite-riot, line 56: Nit: Gradle is a proper noun

nwhite-riot, line 56:
```suggestion
                        <log text="Sentry Gradle plugin will use environment variables to authenticate with Sentry" />
```

plugin-dev/Source/Sentry/Sentry_Android_UPL.xml
```diff
@@ -48,8 +51,16 @@
...
+                    </true>
+                    <false>
+                        <log text="Sentry gradle plugin will use properties file to authenticate with Sentry"/>
```
nwhite-riot, line 59:
```suggestion
                        <log text="Sentry Gradle plugin will use properties file to authenticate with Sentry" />
```

nwhite-riot (approved):

bruno-garcia:
we can't look for the file and fallback for env var? why does it need to be a choice?

tustanivsky:
> we can't look for the file and fallback for env var? why does it need to be a choice?

@bruno-garcia I've updated the scripts so that symbol uploading falls back to env variables if the properties file is missing.

However, it's still worth keeping the new `Configuration type` property and requiring users to set the preferred mechanism explicitly as this makes scripts behavior more predictable for them and provides a better UX in the editor's settings menu.

nwhite-riot:
> However, it's still worth keeping the new Configuration type selector and requiring users to set the preferred mechanism explicitly as this makes scripts behavior more predictable for them

If you provide a fallback, the explicit nature of the configuration type becomes moot. We should either have the setting and not fallback, or remove the setting and handle the fallback.

tustanivsky:
After further consideration based on the points raised, following the current mixed approach with a fallback could lead to unintended side effects. For example, symbol uploads might continue silently with environment variables potentially authenticating to a different project than the one the user expects when relying on the properties file.

I’ve reverted to the original implementation keeping only the explicit configuration type selection in the plugin settings.

bruno-garcia:
I think this is the worst experience at the end. 

It's not a common case that ppl have env vars set to one project but file to another lying around. 

We need to default to a sensible default instead of settings

tustanivsky:
Based on another internal discussion the corresponding selector was removed from the plugin settings leaving only the automatic fallback to environment variables if no properties file is available - #842.

PR #854, opened 2025-03-31
Switch the existing Sentry classes to use USTRUCTs

tustanivsky:


plugin-dev/Source/Sentry/Public/SentryBreadcrumb.h
```diff
@@ -5,60 +5,90 @@
...
 
 /**
  * Information to create a trail of events that happened prior to an issue.
  */
-UCLASS(BlueprintType, NotBlueprintable, HideDropdown)
-class SENTRY_API USentryBreadcrumb : public UObject, public TSentryImplWrapper<ISentryBreadcrumb, USentryBreadcrumb>
+USTRUCT(BlueprintType)
+struct SENTRY_API FSentryBreadcrumb
 {
 	GENERATED_BODY()
```
nwhite-riot, line 20: Should this be `GENERATED_USTRUCT_BODY()`?

plugin-dev/Source/Sentry/Public/SentryBreadcrumb.h
```diff
@@ -5,60 +5,90 @@
...
+	FString GetCategory() const;
+
+	void SetData(const TMap<FString, FString>& Data);
+	TMap<FString, FString> GetData() const;
+
+	void SetLevel(ESentryLevel Level);
+	ESentryLevel GetLevel() const;
+
+	/** Retrieves the underlying native implementation. */
+	TSharedPtr<ISentryBreadcrumb> GetNativeObject() const { return NativeImpl; }
```
nwhite-riot, line 40: In general, it seems odd that we ever expose this. Perhaps we can remove the need for this and any public callers can use a different method? For example, we use `GetNativeObject()` to call `sentry_value_set_stacktrace` - perhaps instead we have a method on the the Sentry event called `AddStacktrace` and that calls the method underneath?

plugin-dev/Source/Sentry/Public/SentryBreadcrumb.h
tustanivsky: This is no longer required nor necessary as `GENERATED_BODY()` is technically the same thing:
```
#define GENERATED_USTRUCT_BODY(...) GENERATED_BODY()
```

PR #853, opened 2025-03-31
Allow editor crashes to be sent to a different sentry project than the packaged game

j0tt:
Hello frens! 

We have our sentry projects set up to isolate crashes which happen in the editor experience from crashes that happen in the cooked title. Since WindowsEngine.ini covers both the editor running on windows as well as the cooked title on windows, there wasn't an out of box way to override the Dsn if in editor (elements in  `EditorConfig.ini` will not override the element in `WindowsConfig.ini`) so I have added a new element to the sentry configuration for `EditorDsn`. Not supplying it simply uses the vanilla `Dsn` element.

I'm opening this pull request to start the conversation around if this is something you would want to take into the official sentry plugin!

PR #859, opened 2025-04-02
Fix `Editor DSN` check on Mac

tustanivsky:
This PR fixes minor compilation error on Mac that appeared after merging #853

#skip-changelog


bitsandfoxes (approved):

j0tt:
Oof, my bad, I misread an existing line of code and thought that would not be needed. Thanks for fixing!

PR #860, opened 2025-04-02
Fix deadlock if ensure event captured during garbage collection

### LINKED ISSUE #850: Reporting ensures during GC causes freeze due to deadlock
### Environment

How do you use Sentry?
Sentry SaaS (sentry.io)

Which version of the SDK?
0.21.1 (but also looks applicable to latest)

Git download

Which version of Unreal?
5.5

Windows desktop

### Steps to Reproduce

We setup our development builds to also report ensure failures to Sentry. When an ensure fails during the garbage collection process, it hangs the game. This can come from any ensure during GC, but in our case it is from AsyncLoading2.cpp

### Expected Result

Should not hang.

### Actual Result

The issue is due to `OnBeforeSend` containing a GC scope guard. This causes sentry to wait for GC to finish, but because the ensure happened during GC, GC never finishes. This causes a mutex deadlock between the two threads. https://github.com/getsentry/sentry-unreal/blob/db5b5afe71a36677aa46a95a558b6aa55f652d50/plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentrySubsystem.cpp#L95

Not sure the right thing to do here, happy to provide more info as needed.

tustanivsky:
Closes #850

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ Blog posts:
 
 - Only crash events captured on Android contain the full callstack. Events that were captured manually won't have the native C++ part there.
 
-- If event was captured during the garbage collection the `BeforeSendHandler` will not be invoked.
+- If an event was captured during the garbage collection, the `BeforeSendHandler` will not be invoked.
 
 - It may be required to upgrade the C++ standard library (`libstdc++`) on older Linux distributions (such as Ubuntu 18.04 and 20.04) to ensure crashpad handler proper functionality within the deployment environment. This can be achieved with something like this:
 ```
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -53,7 +53,7 @@ Blog posts:
 
 - Only crash events captured on Android contain the full callstack. Events that were captured manually won't have the native C++ part there.
 
-- On Windows/Linux if crash event was captured during the garbage collection the `BeforeSendHandler` will not be invoked.
+- If event was captured during the garbage collection the `BeforeSendHandler` will not be invoked.
```
bitsandfoxes, line 56:
```suggestion
- If an event was captured during the garbage collection, the `BeforeSendHandler` will not be invoked.
```

bitsandfoxes (approved):
We should make sure that we're properly documenting the edge-cases where the `beforeSend` does not get invoked. This will trip people up.

PR #864, opened 2025-04-09
chore: Build custom UE Linux Docker images to update CI

### LINKED ISSUE #827: Bump Ubuntu version for `sentry-native` build in CI
GitHub is about to deprecate `ubuntu-20.04` which we're still using in CI to build `sentry-native`:
```
The Ubuntu-20.04 brownout takes place from 2025-02-01. For more details, see https://github.com/actions/runner-images/issues/11101
```
Simply bumping the OS version on the runner where the Native SDK is built breaks the pipeline as it introduces a dependency for `libsentry.a` on `GLIBC_2.34` which's available only in `ubuntu-22.04` and newer. However, the official Linux Docker images for UE 4.27–5.1 provided by Epic Games are based on `ubuntu-18.04` where `GLIBC_2.34` is not available and cannot be installed.

Just like for Windows (#722), we can build custom UE Linux Docker images based on `ubuntu-22.04` allowing us to freely upgrade the OS version on the runner where the plugin's dependencies are built.

tustanivsky:
This PR introduces a new workflow for building custom Unreal Engine Linux Docker images replacing the official ones provided by Epic Games. The official images are still based on `ubuntu-18.04` (for UE 4.27–5.1) which leads to compatibility issues with the `sentry-native` plugin dependency if we update the runner OS version where it's pre-built.

Closes #827

Similar changes were made in #722 while adding Windows support to the existing CI pipeline.

#skip-changelog


.github/workflows/ci.yml
```diff
@@ -152,7 +148,8 @@ jobs:
       fail-fast: false
       matrix:
         # Note: these versions must match scripts/packaging/engine-versions.txt
-        unreal: ['4.27', '5.0', '5.1', '5.2', '5.3', '5.4', '5.5']
+        # Note: Currently 5.0 is not supported due to ue4-docker issue https://github.com/adamrehn/ue4-docker/issues/373
```
bruno-garcia, line 151: It's old enough we could prob remove support btw. Looking at numbers seems to be safe.

Or just call it 'best effort'

.github/workflows/test-linux.yml
```diff
@@ -74,11 +83,11 @@ jobs:
...
-      - name: Setup C++ runtime
-        run: docker exec --user root unreal bash -c '
-          apt-get update ;
-          add-apt-repository ppa:ubuntu-toolchain-r/test ;
-          apt-get install -y libstdc++6'
+      # - name: Setup C++ runtime
+      #   run: docker exec --user root unreal bash -c '
+      #     apt-get update ;
+      #     add-apt-repository ppa:ubuntu-toolchain-r/test ;
+      #     apt-get install -y libstdc++6'
```
bruno-garcia, line 91:
```suggestion
```

.github/workflows/ue-docker-linux.yml
```diff
@@ -0,0 +1,107 @@
...
+on:
+  workflow_dispatch:
+    inputs:
+      ue_version:
+        description: Select Unreal Engine version
+        required: true
+        type: choice
+        options:
+        - 4.27
+        - 5.0
```
bruno-garcia, line 12: Remove?
```suggestion
```

bruno-garcia (approved):

PR #873, opened 2025-04-17
Fix invalid log file being attached to crash events on Mac/iOS

### LINKED ISSUE #302: Crashes captured on Android/iOS have incorrect game log file attached

tustanivsky:
This PR fixes an issue where the game log file from the current session was being attached to a captured crash event before sending it to Sentry instead of attaching the log file from the previous session in which the crash occurred.

It leverages the same approach used for screenshots (#849) by sending the log file in a separate envelope within the `onCrashedLastRun` handler.

Closes #302

Also, supersedes #732


plugin-dev/Source/Sentry/Private/Apple/AppleSentrySubsystem.cpp
```diff
@@ -34,10 +37,12 @@
 #include "Misc/Paths.h"
 #include "UObject/GarbageCollection.h"
 #include "UObject/UObjectThreadContext.h"
-#include "Utils/SentryLogUtils.h"
 
 void FAppleSentrySubsystem::InitWithSettings(const USentrySettings* settings, USentryBeforeSendHandler* beforeSendHandler, USentryBeforeBreadcrumbHandler* beforeBreadcrumbHandler, USentryTraceSampler* traceSampler)
 {
+	isScreenshotAttachmentEnabled = settings->AttachScreenshot;
+	isGameLogAttachmentEnabled = settings->EnableAutoLogAttachment;
```
bitsandfoxes, line 44: Should these go on the actual `options`? We're starting to mix `settings.` with `options.` here. I'd really prefer the `options` to be the source of truth of what options exist and what options the SDK.

tustanivsky: These two flags are cached separately here since they don't necessarily have a direct representation in Cocoa SDK options:

- `isGameLogAttachmentEnabled` is Unreal-specific feature
- `isScreenshotAttachmentEnabled` can't be just set via `options.attachScreenshot` as it's an iOS-only thing in `sentry-coca`

Accessing them using `FSentryModule::Get().GetSettings()` could cause unwanted ensures when called from a non-game thread.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 ### Fixes
 
 - Fix warnings caused by deprecated Cocoa SDK API usages ([#868](https://github.com/getsentry/sentry-unreal/pull/868))
+- Fix invalid log file being attached to crash events on Mac/iOS ([#873](https://github.com/getsentry/sentry-unreal/pull/873))
```
bruno-garcia, line 12: The section on the changelog is already "Fixes" so we could (going forward, from the next release perhaps) avoid starting the message with Fix.

Personally I'd be more interested in which platform the fix relates to. So this is a iOS/Mac, could be:

- iOS/macOS: Correct log file now attached to crashes. Since crashes are sent on restart, the SDK now looks at the previous run's log files to attach. Leaving the current run's log file untouched

Or something like that.

plugin-dev/Source/Sentry/Private/Apple/AppleSentrySubsystem.cpp
bruno-garcia:
```
attachScreenshot
```

This option (with this name) also exists on Android, .NET, Godot, Unity.

plugin-dev/Source/Sentry/Private/Apple/AppleSentrySubsystem.cpp
```diff
@@ -225,17 +233,20 @@ TSharedPtr<ISentryId> FAppleSentrySubsystem::CaptureMessageWithScope(const FStri
...
 }
 
 TSharedPtr<ISentryId> FAppleSentrySubsystem::CaptureEventWithScope(TSharedPtr<ISentryEvent> event, const FSentryScopeDelegate& onConfigureScope)
 {
 	TSharedPtr<SentryEventApple> eventIOS = StaticCastSharedPtr<SentryEventApple>(event);
 
 	SentryId* id = [SENTRY_APPLE_CLASS(SentrySDK) captureEvent:eventIOS->GetNativeObject() withScopeBlock:^(SentryScope* scope) {
+		if(isGameLogAttachmentEnabled) {
+			const FString logFilePath = GetGameLogPath();
+			SentryAttachment* logAttachment = [[SENTRY_APPLE_CLASS(SentryAttachment) alloc] initWithPath:logFilePath.GetNSString()];
```
bruno-garcia, line 247: I'm trying to diff the 3 versions of this snippet in this PR but they all seem the same?

If we're changing things would you want them all to be consistent?
I think it's better to throw this into a function, add some debug logging in there (central point) for example

bruno-garcia (commented):
I'm concerned that we have no tests for this. Do we have iOS device tests we could verify this change and avoid regressions?

plugin-dev/Source/Sentry/Private/Apple/AppleSentrySubsystem.cpp
tustanivsky: I might be missing the concern here. These flags are passed to the Cocoa SDK’s `options` within the `startWithConfigureOptions` block right after we make copies of them here (mainly for convenience as we use those in a few other places beyond initialization). Or is this about the naming?

tustanivsky:
> I'm concerned that we have no tests for this. Do we have iOS device tests we could verify this change and avoid regressions?

At the moment we don’t have a way to automate testing on mobile so manual testing is our only option for now.

bruno-garcia (approved):
Lets ship this to unblock folks, and follow up with device tests if we can

plugin-dev/Source/Sentry/Private/Apple/AppleSentrySubsystem.cpp
tustanivsky: Yes, it's a common naming convention for Unreal - [ISettingsModule::RegisterSettings](https://dev.epicgames.com/documentation/en-us/unreal-engine/API/Developer/Settings/ISettingsModule/RegisterSettings/1)

PR #903, opened 2025-05-08
Fix initializer list parameters order to fix the build on Win/Linux

tustanivsky:
Fixes the build error introduced in #901 

#skip-changelog

### DIFF
--- a/plugin-dev/Source/Sentry/Private/Microsoft/MicrosoftSentrySubsystem.cpp
+++ b/plugin-dev/Source/Sentry/Private/Microsoft/MicrosoftSentrySubsystem.cpp
@@ -9,6 +9,8 @@
 #include "SentrySettings.h"
 
 #include "GenericPlatform/GenericPlatformOutputDevices.h"
+#include "GenericPlatform/GenericPlatformMisc.h"
+#include "Misc/EngineVersionComparison.h"
 
 void FMicrosoftSentrySubsystem::InitWithSettings(
 	const USentrySettings* Settings,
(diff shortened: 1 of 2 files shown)


plugin-dev/Source/Sentry/Private/Microsoft/MicrosoftSentrySubsystem.cpp
```diff
@@ -9,6 +9,8 @@
 #include "SentrySettings.h"
 
 #include "GenericPlatform/GenericPlatformOutputDevices.h"
+#include "GenericPlatform/GenericPlatformMisc.h"
```
nwhite-riot, line 12: Is this header only needed `#if !UE_VERSION_OLDER_THAN(5, 2, 0)`?

tustanivsky: Good catch! I've wrapped that with engine version check as well.

plugin-dev/Source/Sentry/Private/Microsoft/MicrosoftSentrySubsystem.cpp
```diff
@@ -9,6 +9,11 @@
 #include "SentrySettings.h"
 
 #include "GenericPlatform/GenericPlatformOutputDevices.h"
+#include "Misc/EngineVersionComparison.h"
```
nwhite-riot, line 12: Why do you now need this header? Was there a unity issue?

tustanivsky: Yes, it's included explicitly to avoid potential unity-related issues.

bitsandfoxes (approved):

Issue #904, opened 2025-05-08
Provide the expected destination URL for a Sentry event before it is sent


nwhite-riot:
It is often frustrating to dig through recent crashes in Sentry to find your specific crash when you're debugging/reproducing something. If we could make it so that Sentry provided a log line with the expected destination of the Sentry event, it would short circuit all of the waiting.

bruno-garcia:
Having a. link can be tricky because with a `DSN` we don't know what org slug you're in.
But, you can paste the event id and go straight to that event detail in Sentry

![Image](https://github.com/user-attachments/assets/ee547281-a90e-4f24-b788-776f7df5ea73)

@tustanivsky for context:
The event id is a guid defined client side. Often our `CaptureEvent` APIs return the Guid
In some SDks we have `Sentry.LastEventId` which is useful to get the event id from event captured automatically.

PR #905, opened 2025-05-08
Add destination URL for crash envelope

nwhite-riot:
Addresses #904


tustanivsky:
Printing the captured event URL can be handled in the client’s custom `beforeSend` handler where more project-specific context may be available. Doing this in the plugin would require adding an org slug setting, which cannot be determined solely from the provided DSN.

PR #920, opened 2025-05-22
Add API allowing to get/set event fingerprint

tustanivsky:
This PR extends the Event class interface by adding getter and setter functions for the event fingerprint.

Related to https://github.com/getsentry/sentry-docs/issues/13733#event-17680741348


bruno-garcia (approved):
would be nice to have one of the sample errors in the sample project set a fingerprint.

to demonstrate how it works, but also to "test" it when we're running the sample.
Ideally we also test this on device tests/e2e but the sample at least helps verify manually until then

PR #924, opened 2025-05-23
Add instructions how to update plugin package snapshot

### LINKED ISSUE #777: Improve documentation around packaging
As far as I can see, there's no documentation around generating packages, but this step in the contribution process seems required because changing file names/locations etc. will cause the "Check snapshot" job to fail.

1. Update `CONTRIBUTING.md` to include packaging topics (i.e. when to package, what to do, what to commit etc.)
2. Ensure that `pack.ps1` works on Windows with PowerShell 5.1
    a. For example, the `zip` command is not native to Windows
    b. `Out-File` will write UTF8 BOM which causes the file encoding of `SentryModule.cpp` to change unnecessarily

tustanivsky:
This PR adds to `CONTRIBUTING.md` instruction on how to update plugin package snapshot and when it's required.

Closes #777

#skip-changelog

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -45,3 +45,26 @@ In order to run automation tests navigate to `Settings -> Plugins -> Testing` an
 There is a demo level (`SentryDemo.umap`) in project's Content folder which presents a simple UI allowing to send some test events to Sentry. `W_SentryDemo` blueprint implementation shows how to call the plugin API and can be used as a reference.
 
 In order to run the demo level navigate to `Content Browser -> Content -> Maps` and open `SentryDemo` map. Hit play to launch the demo.
+
+## Modifying plugin content
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -45,3 +45,26 @@ In order to run automation tests navigate to `Settings -> Plugins -> Testing` an
...
+```
+
+Once completed, make sure to commit the updated snapshot files to Git.
+
+CI will run a separate check to compare the actual plugin package contents against the snapshot files.
+
+
+
+
+
```
limbonaut, line 70: You may also want to trim these newlines.

limbonaut (approved):
Some abnormal amount of newlines, but otherwise LGTM

CONTRIBUTING.md
tustanivsky: Good catch, thanks!

PR #926, opened 2025-05-24
Remove creator function and replace with HAL include

nwhite-riot:
This allows platform extensions (i.e. console) to define their own `ISentryUser` for use with subsystem `SetUser` etc.

#skip-changelog

### DIFF
--- a/plugin-dev/Source/Sentry/Private/Tests/SentryUser.spec.cpp
+++ b/plugin-dev/Source/Sentry/Private/Tests/SentryUser.spec.cpp
@@ -17,7 +17,7 @@ void SentryUserSpec::Define()
 {
 	BeforeEach([this]()
 	{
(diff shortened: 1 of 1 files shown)


tustanivsky:
Currently `COMPILED_PLATFORM_HEADER` can't be used for Mac/iOS straight away as there we still have a common `Apple` prefix for most platform-specific files (the only exception is `SentrySubsystem` for which we've made separation).

Adding a [link](https://github.com/getsentry/sentry-unreal/pull/779#discussion_r1949012694) to previous conversation regarding this matter for extra context.

nwhite-riot:
@tustanivsky, I've added placeholder IOS and Mac files which include the Apple files - I believe this should resolve the issue, and seems to be how Epic does it in some instances

tustanivsky (approved):

PR #928, opened 2025-05-26
Add native local scope for Windows/Linux

tustanivsky:
This PR adopts the local scope implementation changes for `sentry-native` discussed in https://github.com/getsentry/sentry-native/issues/1226.

Instead of applying scope data to each outgoing event via the `before_send` handler, the SDK now uses a write-only `sentry_scope_t` instance which is then passed to the new `sentry_capture_event_with_scope` function. Also, with this approach global scope can be modified directly via native's API (tags, level, etc.) which addresses the issue with missing context values for fast-fail crashes.

Breaking changes:

- `Dist` and `Environment` properties are now set during the SDK initialization via options, no more corresponding set/get functions in `Scope` class.
- Since [Hub & Scope refactoring docs](https://develop.sentry.dev/sdk/miscellaneous/hub_and_scope_refactoring/) suggest deprecating `ConfigureScope` function and it's a no-op on Win/Linux anyway it was removed from `USentrySubsystem`.

Note: adding breadcrumbs via `sentry_add_breadcrumb` to global scope unblocks further work on https://github.com/getsentry/sentry-native/pull/1166 as custom handling for the corresponding event in Unreal plugin is no longer required.

Docs update - https://github.com/getsentry/sentry-docs/pull/13872


jpnurmi (commented):
looking promising, almost better than i had expected without the getters! 👍

plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentrySubsystem.cpp
```diff
@@ -381,40 +372,44 @@ void FGenericPlatformSentrySubsystem::AddBreadcrumbWithParams(const FString& Mes
...
 
-	scopeStack.Push(NewLocalScope);
-	TSharedPtr<ISentryId> Id = CaptureMessage(message, level);
-	scopeStack.Pop();
+	TSharedPtr<FGenericPlatformSentryScope> NewLocalScope = MakeShareable(new FGenericPlatformSentryScope());
+	onConfigureScope.ExecuteIfBound(NewLocalScope);
+	NewLocalScope->Apply(scope);
 
-	return Id;
+	sentry_uuid_t id = sentry_capture_event_with_scope(nativeEvent, scope);
```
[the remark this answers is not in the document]
jpnurmi: We'll change it to take ownership, after all. If the more complicated scenarios are needed later, we can add another "non-owning" constructor. Sorry for the back and forth!

CHANGELOG.md
```diff
@@ -2,6 +2,15 @@
 
 ## Unreleased
 
+### Breaking Changes
+
+- `Environment` and `Dist` get/set functions were removed from the `Scope` class and now these properties have to be set in plugin settings instead.
+- `ConfigureScope` function was removed from `SentrySubsystem` class.
```
jpnurmi: this is probably ok without deprecating first because it hasn't reached 1.0.0?

jpnurmi (approved):
I'm a bit nervous about the API changes but the sentry-native related changes LGTM 👍

CHANGELOG.md
tustanivsky: Yes, given the scope of the changes in this PR it seems like an appropriate time to remove these APIs ahead of the major version bump.

jpnurmi: Summarizing the reason with a few words and perhaps linking to  https://develop.sentry.dev/sdk/miscellaneous/hub_and_scope_refactoring/ could make the breaking change easier to accept for those who'll be affected 😉

Issue #32, opened 2022-06-22
Support for Gaming Consoles


highlyunavailable:
It would be great if (if possible!) Sentry could provide additional `ISentrySubsystem`s for consoles (e.g. XBox, Playstation, and Switch). This would be a bit more annoying for you because the developer that uses Sentry would have to have the console manufacturer verify that they were a licensed developer to Sentry (E.g. I wold have to have Sony inform you that I'm a licensed PS developer) before you could allow access to that particular SDK but it would be a huge boon for reporting.

bruno-garcia:
This is something we want to look into but we don't have a timeline just yet. Our goal was to get the plugin and support Mobile, Desktop, Web before we get into Consoles.

highlyunavailable:
It's appreciated! I mostly wrote this up as a request so you would know customers are interested in consoles as well.

Edstub207:
This feature will be super useful when it comes, some alternative solutions on the market don't support Console out of the box, so adding this would sway a few folks I think!

Edstub207:
@bruno-garcia / @vaind I don't suppose either of you have an updated timeline for this as the issue is approaching a year old?

highlyunavailable:
@Edstub207 My company is moving off of Sentry (to BugSplat) due to poor Unreal support (bad trace grouping, lack of Linux support for servers) and complete lack of Console support in the foreseeable future, so 🤷‍♀️ . Consider alternatives, IMO.

Edstub207:
> @Edstub207 My company is moving off of Sentry (to BugSplat) due to poor Unreal support (bad trace grouping, lack of Linux support for servers) and complete lack of Console support in the foreseeable future, so 🤷‍♀️ . Consider alternatives, IMO.

Linux crashes seem to be okay for us so far... What do you see?

Edstub207:
@tustanivsky FYI as well - Just realized I've not seen Bruno or Ivan active on this project for a while  :)

Edstub207:
@bruno-garcia @tustanivsky is there any update on this - Had a few other studios wondering about the status of this.

bitsandfoxes:
Hey, sorry about that - Fell through for a bit.
We're looking into it - but it's complicated. (and not just on a technical level)

Edstub207:
> Hey, sorry about that - Fell through for a bit. We're looking into it - but it's complicated. (and not just on a technical level)

[...]

PR #937, opened 2025-05-29
Exclude plugin sources that use sentry-native if build platform isn't supported

tustanivsky:
Related to https://github.com/getsentry/sentry-unreal/issues/32 and https://github.com/getsentry/sentry-docs/issues/13861.

#skip-changelog

### DIFF
--- a/plugin-dev/Source/Sentry/Sentry.Build.cs
+++ b/plugin-dev/Source/Sentry/Sentry.Build.cs
@@ -161,5 +161,14 @@ public class Sentry : ModuleRules
 
 			AddEngineThirdPartyPrivateStaticDependencies(Target, "libcurl");
 		}
+		else
+		{
+			// Exclude sources in `Private/GenericPlatform` that use sentry-native API from the build if target platform isn't supported
+			// Plugin extensions that add support for more platforms (e.g. consoles) can override this define if needed
+			PublicDefinitions.Add("USE_SENTRY_NATIVE=0");
+
+			// TODO: Add a valid link to docs (https://github.com/getsentry/sentry-docs/issues/13861)
+			Console.WriteLine("To use Sentry SDK on game consoles follow the instructions at https://docs.sentry.io/platforms/consoles");
+		}
 	}
 }


plugin-dev/Source/Sentry/Sentry.Build.cs
```diff
@@ -161,5 +161,14 @@ public Sentry(ReadOnlyTargetRules Target) : base(Target)
...
 			AddEngineThirdPartyPrivateStaticDependencies(Target, "libcurl");
 		}
+		else
+		{
+			// Exclude sources in `Private/GenericPlatform` that use sentry-native API from the build if target platform isn't supported
+			// Plugin extensions that add support for more platforms (e.g. consoles) can override this define if needed
+			PublicDefinitions.Add("USE_SENTRY_NATIVE=0");
+
+			// TODO: Add a valid link to docs (https://github.com/getsentry/sentry-docs/issues/13861)
+			Console.WriteLine("To use Sentry SDK on game consoles follow the instructions at https://docs.sentry.io/platforms/consoles");
```
tustanivsky, line 171:
```suggestion
			if (GetType() == typeof(Sentry))
			{
				Console.WriteLine("To use Sentry SDK on game consoles follow the instructions at https://docs.sentry.io/platforms/consoles");
			}
```

tustanivsky: We can print this message to logs only when building for certain platform without the corresponding plugin extension being installed.

Example: for Xbox, this type check should return `Sentry_XSX` if the corresponding extension is installed (see [Sentry_XSX.Build.cs](https://github.com/getsentry/sentry-unreal-xbox/blob/4bc6ea5495d388b6dc0b9dcab43b8c9991094b8c/Sentry/Source/Sentry_XSX.Build.cs)). This indicates that the user has already completed middleware verification and gained access to the `sentry-unreal-xbox` repository so there's no point in printing the link.

bruno-garcia: I think it won't if we're testing type. You'd need to check `IsAssignableTo`

`==` will check for exact type being equal

tustanivsky: I can confirm that `==` checks for the exact type so the message is printed to logs only when there's no corresponding platform extension installed.

tustanivsky:
@nwhite-riot To prevent build errors when packaging for unsupported platforms (e.g., Switch) or when the corresponding extension isn’t installed I’ve also added the `SENTRY_COMPILED_PLATFORM_HEADER` macro (see our previous discussion [here](https://github.com/getsentry/sentry-unreal/pull/779#discussion_r1949012694)).

At some point it might make sense to add platform-specific includes for all Sentry entities so we can use it consistently in `HAL/PlatformSentryXXX.h.` but for now having this in place for `Subsystem` and `User` should be sufficient.

plugin-dev/Source/Sentry/Private/HAL/PlatformSentryDefines.h
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) 2025 Sentry. All Rights Reserved.
+
+#pragma once
+
+// clang-format off
+
+#ifdef SENTRY_PLATFORM_NULL
+	#define SENTRY_COMPILED_PLATFORM_HEADER(Suffix) \
+		PREPROCESSOR_TO_STRING(PREPROCESSOR_JOIN(Null/Null, Suffix))
```
bruno-garcia, line 9: Could we print something like: 

"Sentry supports Unreal Engine on PlayStation, Xbox and Nintendo Switch: https://docs.sentry.io/platforms/unreal/game-consoles/".?

bruno-garcia: oh it's done below. NICE!

bruno-garcia (approved):

PR #940, opened 2025-05-30
Add API allowing to get/set event's `tag`, `context` and `extra` properties

### LINKED ISSUE #936: Capture additional data during BeforeSend
Hello,

We are looking to capture additional data from our Before Send handler - in our case VRAM diagnostics.

The problem is, we can't find a way to do this. 

1. The USentryEvent doesn't have any API to add more data (other than changing the name)
2. The USentryHint is always nullptr on desktop
3. The scope is applied to the event before the before send handler is called, which means adding data to the scope in the handler has no effect on the event

Is there some way we can do this that we aren't seeing? We are using 0.22 currently.

Thanks

tustanivsky:
This PR adds get/set methods to the `USentryEvent` class allowing modification of its tags, contexts, and extras properties. These APIs can be useful when additional data needs to be added to the event during `beforeSend` hook processing.

Closes #936

Related to #951


bruno-garcia:
Part/main reason for BeforeSend is to let users mutate the event, so makes sense

plugin-dev/Source/Sentry/Private/Interface/SentryEventInterface.h
```diff
@@ -20,6 +20,19 @@ class ISentryEvent
 	virtual ESentryLevel GetLevel() const = 0;
 	virtual void SetFingerprint(const TArray<FString>& fingerprint) = 0;
 	virtual TArray<FString> GetFingerprint() = 0;
+	virtual void SetTagValue(const FString& key, const FString& value) = 0;
+	virtual FString GetTagValue(const FString& key) const = 0;
```
limbonaut: Could probably be shorter, without "Value" in the function names. IIRC other SDKs use `set_tag`/`get_tag`.

plugin-dev/Source/Sentry/Private/Tests/SentryEvent.spec.cpp
```diff
@@ -61,6 +61,113 @@ void SentryEventSpec::Define()
...
+			TestEqual("Tags are empty by default", SentryEvent->GetTags().Num(), 0);
+
+			TestEqual("Can't get any tag while these are empty", SentryEvent->GetTagValue(TEXT("TagsKey0")), TEXT(""));
+
+			SentryEvent->SetTagValue(TEXT("TagsKey1"), TEXT("TagsVal1"));
+			SentryEvent->SetTagValue(TEXT("TagsKey2"), TEXT("TagsVal2"));
+
+			TestEqual("There are two tags adding them as individual items", SentryEvent->GetTags().Num(), 2);
+
+			TestEqual("First tag persist its original value", SentryEvent->GetTagValue(TEXT("TagsKey1")), TEXT("TagsVal1"));
```
limbonaut:
```suggestion
			TestEqual("First tag retains its original value", SentryEvent->GetTagValue(TEXT("TagsKey1")), TEXT("TagsVal1"));
```
nitpick: description could be more clear (and in similar cases below)

limbonaut (approved):
Looks good! Covered by tests -- that's awesome. Just a few nitpicks.

plugin-dev/Source/Sentry/Private/Apple/AppleSentryEvent.cpp
```diff
@@ -66,6 +66,83 @@ TArray<FString> FAppleSentryEvent::GetFingerprint()
 	return FAppleSentryConverters::StringArrayToUnreal(EventApple.fingerprint);
 }
 
+void FAppleSentryEvent::SetTagValue(const FString& key, const FString& value)
+{
+	NSMutableDictionary* mutableTags = [EventApple.tags mutableCopy] ?: [NSMutableDictionary dictionary];
```
bruno-garcia, line 71: if we make copies here, do we need to copy below again?
can we keep this new copy instead?

plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentryEvent.cpp
```diff
@@ -75,6 +75,138 @@ TArray<FString> FGenericPlatformSentryEvent::GetFingerprint()
...
+	}
+	else
+	{
+		sentry_value_set_by_key(eventContexts, TCHAR_TO_ANSI(*key), FGenericPlatformSentryConverters::StringMapToNative(values));
+	}
+}
+
+TMap<FString, FString> FGenericPlatformSentryEvent::GetContext(const FString& key) const
+{
+	sentry_value_t eventContexts = sentry_value_get_by_key(Event, "contexts");
```
bruno-garcia, line 162: TValue is a string because it's always an object pre-serialized? sentry doesn't accept string as a value for contexts, it has to be an object.

plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentryEvent.cpp
```diff
@@ -75,6 +75,138 @@ TArray<FString> FGenericPlatformSentryEvent::GetFingerprint()
...
+		sentry_value_set_by_key(eventExtra, TCHAR_TO_ANSI(*key), sentry_value_new_string(TCHAR_TO_ANSI(*value)));
+	}
+}
+
+FString FGenericPlatformSentryEvent::GetExtraValue(const FString& key) const
+{
+	sentry_value_t eventExtra = sentry_value_get_by_key(Event, "extra");
+	if (sentry_value_is_null(eventExtra))
+	{
+		return FString();
```
bruno-garcia: should we return nullptr instead of these empty strings? 

"" is a value and tells me the key existed.

plugin-dev/Source/Sentry/Private/Tests/SentryEvent.spec.cpp
```diff
@@ -61,6 +61,113 @@ void SentryEventSpec::Define()
...
 		});
 	});
+
+	Describe("Event tags", [this]()
+	{
+		It("can be added and removed", [this]()
+		{
+			TestEqual("Tags are empty by default", SentryEvent->GetTags().Num(), 0);
+
+			TestEqual("Can't get any tag while these are empty", SentryEvent->GetTagValue(TEXT("TagsKey0")), TEXT(""));
```
bruno-garcia: I'd expect null instead

plugin-dev/Source/Sentry/Private/Tests/SentryEvent.spec.cpp
```diff
@@ -61,6 +61,113 @@ void SentryEventSpec::Define()
...
+			TestEqual("There are no tags after setting an empty map", SentryEvent->GetTags().Num(), 0);
+		});
+	});
+
+	Describe("Event contexts", [this]()
+	{
+		It("can be added and removed", [this]()
+		{
+			TMap<FString, FString> InTestContext;
+			InTestContext.Add(TEXT("ContextKey1"), TEXT("ContextVal1"));
```
bruno-garcia, line 122: this will break in sentry I think. it expects an object not a string

tustanivsky: `InTestContext` var is a string map that represents context values, not the context itself. Internally SDK converts it to an object and assigns to the context with specified key ("TestContext1" in this case).

plugin-dev/Source/Sentry/Private/Interface/SentryEventInterface.h
tustanivsky: Makes sense, I've updated the naming.

plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentryEvent.cpp
tustanivsky: Not sure I understand. Here `eventContexts` is an object that stores all event's contexts. We use it to retrieve one specific context (which is also an object) with given key. This context object is then converted to a string map that we return to Unreal.

plugin-dev/Source/Sentry/Private/Apple/AppleSentryEvent.cpp
tustanivsky: Yes, we can skip making extra copies here.

plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentryEvent.cpp
tustanivsky: Currently, returning `FString()` is the only blueprint-compatible way to signal a missing value unless we add a separate `bFound` output or wrap it in some struct.

tustanivsky: Opened a separate ticket for this - https://github.com/getsentry/sentry-unreal/issues/951

limbonaut: If the optional/nullable is not available, perhaps adding `HasTag(key)` and `HasExtra(key)` would be a viable alternative?

plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentryEvent.cpp
limbonaut: I think I understand. In this context, Bruno probably means that the context values are not necessary strings. It can be other types too, including a map (aka object).
In other words, it should be something like `TMap<FString, TVariant>`. Any possible [`sentry_value_type_t`](https://github.com/getsentry/sentry-native/blob/70982f57b6407a1f3afe377d0c2a61ccbbfb2ce2/include/sentry.h#L136-L144) is a valid value inside a context. 

EDIT: Or maybe it's a different point -- totally missed it in my review.

plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentryEvent.cpp
tustanivsky: Yes, this could be a solution as well - thanks for bringing it up!

plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentryEvent.cpp
tustanivsky: Ah, now I see. Currently, the Unreal plugin only allows setting context values as strings so we're essentially expecting them to be strings when retrieving context as well.

limbonaut: Here's an example:
Screenshot 2025-06-06 at 11 38 20
The "display" context has a key called "screens", the value of which is a list, and the first element of that list is a map (aka object). So this function should be able to support such data structure.

limbonaut: > Currently, the Unreal plugin only allows setting context values as strings so we're essentially expecting them to be strings when retrieving context as well.

So, if a user does GetContext() followed by SetContext() with a `device` context, the values will get converted to all strings in this case, right? It can be a problem.

tustanivsky: Right, I overlooked the case of non-user-defined contexts that may have a more complex structure. I'll move the PR back to draft to address this.

plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentryEvent.cpp
nwhite-riot: Is `TOptional<...>` not available to blueprints? 

Alternatively, can blueprints have an out-parameter exposed?
```
bool TryGetExtraValue(const FString& Key, FString& OutValue)
```

plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentryEvent.cpp
nwhite-riot: Would it make sense to store JSON objects in the `TMap<..., ...>` instead of trying to come up with a variant structure?

plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentryEvent.cpp
tustanivsky: > Is TOptional<...> not available to blueprints?

Yes, we can't use `TOptional` for APIs that are exposed to blueprints.

> Alternatively, can blueprints have an out-parameter exposed?

That makes sense - providing getters with out-parameters sounds like a good way to move forward with this.

plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentryEvent.cpp
tustanivsky: > Would it make sense to store JSON objects in the TMap<..., ...> instead of trying to come up with a variant structure?

Afaik `FJsonObject` can't be exposed to blueprints as well.

I've already added a basic implementation of [FSentryVariant](https://github.com/getsentry/sentry-unreal/blob/8b8f7bdddf9480029b3166d4f7d3d2fa6661f2d2/plugin-dev/Source/Sentry/Public/SentryVariant.h) (built on top of `FVariant`) that can be used both in C++ and blueprints. If `TMap<FString, FSentryVariant>` proves to be a viable replacement for the current approach we should be good to stick with it.

plugin-dev/Source/Sentry/Private/Tests/SentryEvent.spec.cpp
bruno-garcia: I see, so I could add

```
InTestContext.Add(TEXT("ContextKey1"), TEXT("ContextVal1"));
InTestContext.Add(TEXT("ContextKey1"), TEXT("ContextVal2"));
```


What would it be?

```json
{ 
"ContextKey1": {
"??":"ContextVal1",
"??":"ContextVal2",
}}
```

tustanivsky: No, adding this values just modifies the context called `TestContext1`, so the result will look like this:
```json
{
  "contexts":
  {
    "TestContext1":
    {
      "ContextKey1": "ContextVal1",
      "ContextKey2": "ContextVal2"
    }
  }
}
```

tustanivsky:
To address the concerns raised in the [discussion](https://github.com/getsentry/sentry-unreal/pull/940#discussion_r2128781887) above I've added [FSentryVariant](https://github.com/getsentry/sentry-unreal/blob/7b5891993694bbccd6cb19440e598c93233a32f3/plugin-dev/Source/Sentry/Public/SentryVariant.h) which allows populating an event's context with values of various data types (previously only `FString` was supported) and properly handle contexts with complex structures (e.g., those containing nested objects):

Screenshot 2025-06-12 162924

Unlike default `FVariant` it supports both in C++ and blueprints. `USentryVariantHelper` exposes number of convenience functions for blueprint users so that they can use `FSentryVariant` pretty much like any other engine type there:

image

To be consumed by the underlying SDKs the variant is converted to a platform-specific representation for Native, Android and Cocoa.

Tests for common variant use cases are located in [SentryVariant.spec.cpp](https://github.com/getsentry/sentry-unreal/blob/7b5891993694bbccd6cb19440e598c93233a32f3/plugin-dev/Source/Sentry/Private/Tests/SentryVariant.spec.cpp).

plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentryEvent.cpp
tustanivsky: Added `TryGetTag`, `TryGetContext` and `TryGetExtraValue` functions to the event's API in order to address this.

plugin-dev/Source/Sentry/Private/Tests/SentryEvent.spec.cpp
tustanivsky: Added `TryGetTag` function allowing to check whether a specific tag exists. Unlike `GetTag`, which returns an empty string if the tag is not found, `TryGetTag` explicitly indicates presence via its return value (see above [discussion](https://github.com/getsentry/sentry-unreal/pull/940#discussion_r2128785430))

plugin-dev/Source/Sentry/Private/Android/AndroidSentryEvent.cpp
```diff
@@ -70,6 +80,132 @@ TArray<FString> FAndroidSentryEvent::GetFingerprint()
...
+	return true;
+}
+
+void FAndroidSentryEvent::RemoveContext(const FString& key)
+{
+	FSentryJavaObjectWrapper::CallStaticMethod<void>(SentryJavaClasses::SentryBridgeJava, "removeContext", "(Lio/sentry/SentryEvent;Ljava/lang/String;)V",
+		GetJObject(), *FSentryJavaObjectWrapper::GetJString(key));
+}
+
+void FAndroidSentryEvent::SetExtraValue(const FString& key, const FString& value)
```
limbonaut:
```suggestion
void FAndroidSentryEvent::SetExtraValue(const FString& key, const FSentryVariant& value)
```
Should probably also accept the variant (and in other "extra" functions).

plugin-dev/Source/Sentry/Private/Android/Infrastructure/AndroidSentryJavaObjectWrapper.cpp
```diff
@@ -100,6 +100,25 @@ FScopedJavaObject<jstring> FSentryJavaObjectWrapper::GetJString(const FString& S
...
+	jclass ClassGlobalRef;
+
+	if (ClassData.Type == ESentryJavaClassType::System)
+		ClassGlobalRef = FJavaWrapper::FindClassGlobalRef(JEnv, AnsiClassName, false);
+	if (ClassData.Type == ESentryJavaClassType::External)
+		ClassGlobalRef = AndroidJavaEnv::FindJavaClassGlobalRef(AnsiClassName);
+
+	check(ClassGlobalRef);
+
+	return JEnv->IsInstanceOf(JavaClassInstance, ClassGlobalRef);
```
limbonaut, line 119: It's probably a good idea to cache these search results, as this method is going to be called many times from the "context" methods and converters. Maybe cache it in the class data?

plugin-dev/Source/Sentry/Private/Apple/Infrastructure/AppleSentryConverters.cpp
```diff
@@ -147,6 +192,68 @@ TArray<uint8> FAppleSentryConverters::ByteDataToUnreal(NSData* data)
...
+		NSNumber* number = (NSNumber*)variant;
+
+		const char* objCType = [number objCType];
+
+		// NSNumber created from bool can be encoded as char
+		if (strcmp(objCType, @encode(bool)) == 0 || strcmp(objCType, @encode(char)) == 0)
+		{
+			return FSentryVariant([number boolValue]);
+		}
+		if (strcmp(objCType, @encode(float)) == 0)
```
limbonaut, line 208: Q: Should we also add conversion for `double` and `long`?

plugin-dev/Source/Sentry/Public/SentryVariant.h
```diff
@@ -0,0 +1,254 @@
...
+	GENERATED_BODY()
+
+	FSentryVariant();
+	FSentryVariant(int32 InValue);
+	FSentryVariant(float InValue);
+	FSentryVariant(bool InValue);
+	FSentryVariant(const FString& InValue);
+	FSentryVariant(const TCHAR* InValue);
+	FSentryVariant(const TArray<FSentryVariant>& InValue);
+	FSentryVariant(const TMap<FString, FSentryVariant>& InValue);
```
limbonaut, line 52: You could also add conversion operators like `operator float()` to make assignments easier, like `float x = myVariant`.

plugin-dev/Source/Sentry/Public/SentryEvent.h
```diff
@@ -51,6 +52,70 @@ class SENTRY_API USentryEvent : public UObject, public TSentryImplWrapper<ISentr
...
+	UFUNCTION(BlueprintCallable, Category = "Sentry")
+	void SetExtraValue(const FString& Key, const FString& Value);
+
+	/** Gets an extra of the event. */
+	UFUNCTION(BlueprintPure, Category = "Sentry")
+	FString GetExtraValue(const FString& Key) const;
+
+	/** Tries to get an extra of the event. */
+	UFUNCTION(BlueprintPure, Category = "Sentry")
+	bool TryGetExtraValue(const FString& Key, FString& Value) const;
```
jpnurmi: could the "Value" suffix be dropped for these as well to make them consistent with the others?

plugin-dev/Source/Sentry/Private/Android/AndroidSentryEvent.cpp
tustanivsky: Right, I've replaced strings with variants for all "extra" functions

plugin-dev/Source/Sentry/Public/SentryEvent.h
tustanivsky: The `Value` suffix is now removed.

plugin-dev/Source/Sentry/Private/Apple/Infrastructure/AppleSentryConverters.cpp
tustanivsky: I suppose we don't need conversions for `double` and `long` at this point since these types aren't available in blueprints and thus `FSentryVariant` doesn't support them.

plugin-dev/Source/Sentry/Public/SentryVariant.h
tustanivsky: [Done](https://github.com/getsentry/sentry-unreal/pull/940/commits/b2b6e9236ab60831fa26610f59f429c2bb3757f0)!

plugin-dev/Source/Sentry/Private/Android/Infrastructure/AndroidSentryJavaObjectWrapper.cpp
tustanivsky: @limbonaut I've managed to come up with a solution allowing to cache `jclass` references to avoid extra look-ups however it feels like something that would be better to address in a separate ticket. The current `AndroidSentryJavaObjectWrapper` implementation could benefit from broader refactoring and optimization which I'd prefer not to include in this PR. Since it's a non-blocking issue are you okay with moving forward as is?

plugin-dev/Source/Sentry/Private/Android/Infrastructure/AndroidSentryConverters.cpp
```diff
@@ -202,6 +257,95 @@ TArray<uint8> FAndroidSentryConverters::ByteArrayToUnreal(jbyteArray byteArray)
...
+	if (FSentryJavaObjectWrapper::IsInstanceOf(SentryJavaClasses::List, variant))
+	{
+		return FSentryVariant(FAndroidSentryConverters::VariantArrayToUnreal(variant));
+	}
+	if (FSentryJavaObjectWrapper::IsInstanceOf(SentryJavaClasses::Map, variant))
+	{
+		return FSentryVariant(FAndroidSentryConverters::VariantMapToUnreal(variant));
+	}
+
+	return FSentryVariant();
```
bruno-garcia, line 296: do we need a debug/diagnostic log here? I imagine this could be hard to debug if this manifests of missing data and we dunno why

tustanivsky: Makes sense - added log message here

PR #953, opened 2025-06-06
Add option to delay app shutdown until Crashpad completes crash report upload

### LINKED ISSUE #914: When running on docker with crashpad wait for crashpad handler
More context here:
* https://github.com/getsentry/sentry-native/issues/1237

Alternatively instead of baking this setting in (since we can't be sure it's running on docker), we could have a setting for the user to select, and a note on the docs.

tustanivsky:
This PR adds a new option to the plugin settings that allows delaying app shutdown until the crash report is uploaded. This can be useful when running an Unreal game in a Docker environment (Windows/Linux) where containers typically don’t persist their state causing crash reports to be lost before they are uploaded to Sentry.

The feature was introduced to Natives SDK in https://github.com/getsentry/sentry-native/pull/1255

Docs update - https://github.com/getsentry/sentry-docs/pull/13946

Closes #914


jpnurmi (approved):
~~nit-pick: should it mention in the docs or property description that it applies to Windows and Linux only?~~

edit: oh wait, sentry-unreal doesn't even use sentry-native on macos? ignore this! :man_facepalming:

limbonaut (approved):

plugin-dev/Source/Sentry/Public/SentrySettings.h
```diff
@@ -285,6 +285,10 @@ class SENTRY_API USentrySettings : public UObject
 		Meta = (DisplayName = "Sentry database location", ToolTip = "Location where Sentry stores its internal/temporary files."))
 	ESentryDatabaseLocation DatabaseLocation;
 
+	UPROPERTY(Config, EditAnywhere, Category = "General|Native",
```
bruno-garcia, line 288: Should we have this `true` by default?
If we can detect it's running on docker ideally, perhaps there are some heuristics?

I mean Linux is mostly for servers (outside SteamOS?) and Windows if it's Windows Server, it's also fine to be `true` by default.

tustanivsky: Currently we’re just using sentry-native's default here which is `false`.

Since running UE app in a Docker container is a relatively specific use case it may be worth to keep this opt-in for now (at least until we receive some feedback from users who had raised this issue earlier).

bruno-garcia (approved):

PR #967, opened 2025-06-18
chore: Update Windows runners for pre-building Native SDK in CI

### LINKED ISSUE #966: Update Windows runners for pre-building `sentry-native` in CI
Currently, CI uses `windows-2019` runners to pre-build the Native SDK, a key dependency of the Unreal plugin. These runners include `Visual Studio 2019` which is required to ensure compatibility with older engine versions.

GitHub is about to deprecate `windows-2019` runners so CI has to be updated accordingly.

```
The windows-2019 runner image is being deprecated, consider switching to windows-2022(windows-latest) or windows-2025 instead. For more details see https://github.com/actions/runner-images/issues/12045.
```

tustanivsky:
Closes #966 

#skip-changelog

### DIFF
--- a/scripts/build-win64-crashpad.sh
+++ b/scripts/build-win64-crashpad.sh
@@ -6,6 +6,7 @@ export sentryArtifactsDestination=$2
 
 rm -rf "${sentryArtifactsDestination}/"*
 
+# Force CMake to use the VS2019 toolset (`-T v142`) to maintain native libraries compatibility with older Unreal Engine versions (4.27–5.1)
 cmake -G "Visual Studio 17 2022" -T v142 -S "${sentryNativeRoot}" -B "${sentryNativeRoot}/build" -D SENTRY_BACKEND=crashpad -D SENTRY_SDK_NAME=sentry.native.unreal -D SENTRY_BUILD_SHARED_LIBS=OFF
 cmake --build "${sentryNativeRoot}/build" --target sentry --config RelWithDebInfo --parallel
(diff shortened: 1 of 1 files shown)


scripts/build-win64-breakpad.sh
```diff
@@ -6,7 +6,7 @@ export sentryArtifactsDestination=$2
 
 rm -rf "${sentryArtifactsDestination}/"*
 
-cmake -G "Visual Studio 16 2019" -S "${sentryNativeRoot}" -B "${sentryNativeRoot}/build" -D SENTRY_BACKEND=breakpad -D SENTRY_SDK_NAME=sentry.native.unreal -D SENTRY_BUILD_SHARED_LIBS=OFF
+cmake -G "Visual Studio 17 2022" -T v142 -S "${sentryNativeRoot}" -B "${sentryNativeRoot}/build" -D SENTRY_BACKEND=breakpad -D SENTRY_SDK_NAME=sentry.native.unreal -D SENTRY_BUILD_SHARED_LIBS=OFF
```
limbonaut, line 10: Might be worth leaving a comment that `-T v142` means using 2019 toolset

limbonaut (approved):

scripts/build-win64-breakpad.sh
tustanivsky: That makes sense, thanks - the corresponding comment was added

jpnurmi (approved):

PR #969, opened 2025-06-20
Remove BOM from source files

yangskyboxlabs:
BOM causes Perforce's encoding detection to flag the file as "utf8" instead of "text" which can be undesirable.  Remove these BOMs to make to mitigate getting into this particular niche issue.


tustanivsky:
Linking #970 for extra context.

tustanivsky (approved):

Issue #951, opened 2025-06-06
Improve handling of missing string values in Blueprint-exposed getter functions


tustanivsky:
Getter functions for tags, contexts, extras, etc. return `FString()` to indicate missing values due to Blueprint exposure limitations (can't use `nullptr` or `TOptional`). This can be ambiguous as `""` might also be a valid value so there's a need to investigate alternatives such as adding a `bFound` out parameter or using a custom struct to clearly indicate whether a value was found. Exposing additional APIs like `HasTag`, `HasConext`, etc. could be a viable solution as well as providing getters that use out-parameters e.g. `bool TryGetExtraValue(const FString& Key, FString& OutValue)`.

Internal wrappers for platform-specific code still can use `nullptr` or `TOptional` which can be a good starting point for this improvement.

Originally this question was raised in [comment](https://github.com/getsentry/sentry-unreal/pull/940#discussion_r2128785430) for #940.

PR #971, opened 2025-06-20
Adopt generic variant type in public APIs

tustanivsky:
This PR adopts the `FSentryVariant` type introduced in #940 for use in public APIs that handle arbitrary Sentry objects data (e.g. contexts and extras in `USentryScope`, data in `USentryBreadcrumb`). This allows to move away from using `FString` in favor of a more flexible and type-safe approach.

It also adds TryGet-functions for scope's tags, extras and contexts (#951).


tustanivsky:
On Android, scope's extras that are variants with array/map values will be set as strings due to current `sentry-java` limitations (see https://github.com/getsentry/sentry-java/issues/2032):

- Example `extras` when added via `sentry-native` on desktop
image

- Example `extras` when added via `sentry-java` on Android
image

plugin-dev/Source/Sentry/Private/GenericPlatform/Infrastructure/GenericPlatformSentryConverters.cpp
```diff
@@ -197,50 +197,56 @@ FSentryVariant FGenericPlatformSentryConverters::VariantToUnreal(sentry_value_t
...
-		for (int32 i = 0; i < len; ++i)
-		{
-			unrealArray.Add(VariantToUnreal(sentry_value_get_by_index(variant, i)));
-		}
+TMap<FString, FSentryVariant> FGenericPlatformSentryConverters::VariantMapToUnreal(sentry_value_t map)
+{
+	TMap<FString, FSentryVariant> unrealMap;
 
-		return FSentryVariant(unrealArray);
+	FString mapJsonString = FString(sentry_value_to_json(map));
```
jpnurmi: `sentry_value_to_json` returns a newly allocated string that must be freed but feel free to fix it in a separate PR because this is not new code and this PR is already quite large :)

sample/Content/UI/W_SentryDemo.uasset
jpnurmi, line 1: slipped in by accident? :)

tustanivsky: Not really - this is a blueprint representing the demo app widget which needed some adjustments after the API changes.

plugin-dev/Source/Sentry/Private/GenericPlatform/Infrastructure/GenericPlatformSentryConverters.cpp
tustanivsky: Good catch! I've added missing `sentry_string_free` calls.

PR #982, opened 2025-06-26
Add runtime attachments support for Windows/Linux

### LINKED ISSUE #430: Enable Attachments in Windows

https://github.com/getsentry/sentry-unreal/blob/main/plugin-dev/Source/Sentry/Private/Desktop/SentryScopeDesktop.cpp#L34-L37 

This is a request from a customer:

Adding support for this would enable "...custom filtered log files for particular subsystems, .utrace files, and any other artifacts we might want to upload via a CaptureMessageWithScope call."

### LINKED ISSUE #946: FGenericPlatformSentryScope::AddAttachment not implemented
Working with the Sentry plugin in an Unreal Engine project, I would like to be able to attach a text file in sentry event, but `void FGenericPlatformSentryScope::AddAttachment(TSharedPtr<ISentryAttachment> attachment)` is not implemented. I've found a work around by adding a global `sentry_options_add_attachmentw` call to attach the file for all Sentry events, and this seems to solve the problem, but wondering if there's any plan to support `AddAttachment` in the future?

tustanivsky:
This PR adopts changes introduced in Native SDK version [0.9.1](https://github.com/getsentry/sentry-native/releases/tag/0.9.1) to enable runtime attachments on Windows and Linux.

Attachments can now be added to all outgoing events by setting them via `USentrySubsystem` on the global scope or to individual events by adding them to a local scope, i.e. when capturing with `CaptureEventWithScope`.

On Apple/Android, removing individual attachments from the global scope currently is not supported. However, on these platforms the corresponding SDKs provide `clearAttachments` which is not available in `sentry-native` at the moment.

Closes #430
Closes #946

### DIFF
--- a/plugin-dev/Source/Sentry/Private/Android/Java/SentryBridgeJava.java
+++ b/plugin-dev/Source/Sentry/Private/Android/Java/SentryBridgeJava.java
@@ -232,6 +232,6 @@ public class SentryBridgeJava {
 	}
 
 	public static void removeAttachment(final Attachment attachment) {
-		Sentry.getGlobalScope().getAttachments().remove(attachment);
+		// Currently, Android SDK doesn't have API allowing to remove individual attachments
 	}
 }
--- a/plugin-dev/Source/Sentry/Private/Android/Java/SentryBridgeJava.java
+++ b/plugin-dev/Source/Sentry/Private/Android/Java/SentryBridgeJava.java
@@ -14,6 +14,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 
+import io.sentry.Attachment;
 import io.sentry.Breadcrumb;
 import io.sentry.Hint;
 import io.sentry.IScopes;
@@ -224,5 +225,13 @@ public class SentryBridgeJava {
 
 	public static void setScopeExtra(final IScope scope, final String key, final Object values) {
 		scope.setExtra(key, values.toString());
+	}
+
+	public static void addAttachment(final Attachment attachment) {
+		Sentry.getGlobalScope().addAttachment(attachment);
+	}
+
+	public static void removeAttachment(final Attachment attachment) {
+		Sentry.getGlobalScope().getAttachments().remove(attachment);
 	}
 }
--- a/plugin-dev/Source/Sentry/Private/Android/Java/SentryBridgeJava.java
+++ b/plugin-dev/Source/Sentry/Private/Android/Java/SentryBridgeJava.java
@@ -234,4 +234,9 @@ public class SentryBridgeJava {
 	public static void removeAttachment(final Attachment attachment) {
 		// Currently, Android SDK doesn't have API allowing to remove individual attachments
 	}
+
+	public static void clearAttachments() {
+		Sentry.getGlobalScope().clearAttachments();
+	}
+
 }
--- a/plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentrySubsystem.cpp
+++ b/plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentrySubsystem.cpp
@@ -220,6 +220,8 @@ void FGenericPlatformSentrySubsystem::AddFileAttachment(TSharedPtr<ISentryAttach
 		sentry_attachment_set_content_type(nativeAttachment, TCHAR_TO_UTF8(*platformAttachment->GetContentType()));
 
 	platformAttachment->SetNativeObject(nativeAttachment);
+
+	attachments.Add(platformAttachment);
 }
 
 void FGenericPlatformSentrySubsystem::AddByteAttachment(TSharedPtr<ISentryAttachment> attachment)
@@ -235,6 +237,8 @@ void FGenericPlatformSentrySubsystem::AddByteAttachment(TSharedPtr<ISentryAttach
 		sentry_attachment_set_content_type(nativeAttachment, TCHAR_TO_UTF8(*platformAttachment->GetContentType()));
 
 	platformAttachment->SetNativeObject(nativeAttachment);
+
+	attachments.Add(platformAttachment);
 }
 
 FGenericPlatformSentrySubsystem::FGenericPlatformSentrySubsystem()
@@ -437,6 +441,16 @@ void FGenericPlatformSentrySubsystem::RemoveAttachment(TSharedPtr<ISentryAttachm
 	sentry_remove_attachment(nativeAttachment);
 
 	platformAttachment->SetNativeObject(nullptr);
+}
+
+void FGenericPlatformSentrySubsystem::ClearAttachments()
+{
+	for (auto& attachment : attachments)
+	{
+		RemoveAttachment(attachment);
+	}
+
+	attachments.Empty();
 }
 
 TSharedPtr<ISentryId> FGenericPlatformSentrySubsystem::CaptureMessage(const FString& message, ESentryLevel level)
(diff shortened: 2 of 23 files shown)


plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentrySubsystem.cpp
```diff
@@ -381,6 +413,27 @@ void FGenericPlatformSentrySubsystem::ClearBreadcrumbs()
...
+	}
+}
+
+void FGenericPlatformSentrySubsystem::RemoveAttachment(TSharedPtr<ISentryAttachment> attachment)
+{
+	TSharedPtr<FGenericPlatformSentryAttachment> platformAttachment = StaticCastSharedPtr<FGenericPlatformSentryAttachment>(attachment);
+
+	sentry_attachment_t* nativeAttachment = platformAttachment->GetNativeObject();
+
+	sentry_remove_attachment(nativeAttachment);
```
jpnurmi, line 441: If I understand correctly, `FGenericPlatformSentryAttachment` outlives the `sentry_native_t` instance here. Maybe we could reset the native object reference just to be safe and avoid accidentally dereferencing a dangling pointer somewhere down the line?

tustanivsky: Fair enough, I’ve added a reference reset here and a default initializer to ensure we don’t end up with an invalid pointer at some point.

tustanivsky:
> There might be some unintentional copies when passing the byte buffer around:
> 
> ```c++
> const TArray<uint8>& byteBuf = attachment->GetData();
> ...
> ```
> 
> The reference makes it look like no copy is happening but `GetData()` returns `TArray<uint8>` by value. Since it returns a member variable and not a temporary, RVO won't kick in either. If it's ok for the internal platform interface to return by reference, that would save us some copies.

@jpnurmi On [Android](https://github.com/getsentry/sentry-unreal/blob/3711a9f1e8bab2745cce3462258d85485f480d3e/plugin-dev/Source/Sentry/Private/Android/AndroidSentryAttachment.cpp#L33) and [iOS](https://github.com/getsentry/sentry-unreal/blob/3711a9f1e8bab2745cce3462258d85485f480d3e/plugin-dev/Source/Sentry/Private/Apple/AppleSentryAttachment.cpp#L35) we're retrieving attachment's byte data from the corresponding native objects via conversion utils so we can't really return it by reference as result is a temporary.

However, we can introduce a `GetDataByRef()` method specifically for `FGenericPlatformSentryAttachment` without changing the platform interface. This would avoid unnecessary copying of byte arrays when adding them to the scope. If the user calls `GetData()` a copy will still be made but in that case the API makes it explicit. Does that make sense?

plugin-dev/Source/Sentry/Private/Android/Java/SentryBridgeJava.java
```diff
@@ -225,4 +226,12 @@ public static Object getScopeContext(final IScope scope, final String key) {
 	public static void setScopeExtra(final IScope scope, final String key, final Object values) {
 		scope.setExtra(key, values.toString());
 	}
+
+	public static void addAttachment(final Attachment attachment) {
+		Sentry.getGlobalScope().addAttachment(attachment);
+	}
+
+	public static void removeAttachment(final Attachment attachment) {
+		// Currently, Android SDK doesn't have API allowing to remove individual attachments
```
limbonaut, line 235: This may become a gotcha moment for users. Maybe this should be raised internally. Is it necessary to have the `removeAttachment` method on a global scope? If so, should it also be included in the underlying SDKs?
Same issue in Godot SDK: https://github.com/getsentry/sentry-godot/pull/205#discussion_r2166510716

tustanivsky: Given the discrepancies between Native, Android, and Cocoa in terms of attachment API support it looks like we have the following options:

1. Add `clearAttachments` method to Native which is already available for Android/Cocoa and expose that for global scope instead of giving user the ability to remove individual attachments
2. Add `removeAttachment` to Android/Cocoa to enable removing individual attachments on mobile
3. Combine 1 and 2 to cover all possible use cases
4. Leave only `addAttachment` for global scope without the ability to remove it
5. Leave things as-is and document the current limitations

jpnurmi: I don't see any problems with adding a new sentry-native method to clear attachments in the global scope.

tustanivsky: Sounds good - we can expose only the `AddAttachment` and `ClearAttachments` methods for the global scope to align with the mobile SDKs.

This could also simplify the `FGenericPlatformSentryAttachment` implementation as we’d no longer need to store a `sentry_attachment_t` reference there.

limbonaut: Wouldn't clear_attachments() also nuke a screenshot attachment? I wonder if it's going to be an unfortunate side-effect.

jpnurmi: I noticed a similar undesired side-effect when telling Crashpad to clear all attachments. It accidentally cleared some internal `__sentry-event` and `__sentry-breadcrumb` attachments that must not be removed or Sentry won't receive the information it needs...

tustanivsky: Maybe we can clear only attachments that were added by user manually via the API? I mean without touching the internal stuff and those that were added during the init using `sentry_options_add_attachment`

jpnurmi: The problem kind of bubbles up. Crashpad doesn't know which attachments are special from sentry-native's perspective. Likewise, sentry-native might not know which attachments are special from a downstream SDKs perspective.

I like the idea of making a distinction between fixed attachments added via options vs. dynamic attachments added via the new API, even though during sentry-native attachment API reviews it was proposed that the old options API could be removed when the missing Crashpad IPC was implemented for macOS. Not sure if it's going to happen anytime soon for macOS using rather complicated Mach ports for IPC, though. :sweat_smile:

tustanivsky: For now, I’ve worked around this by implementing `ClearAttachments` for Win/Linux on the Unreal side of things (see [commit](https://github.com/getsentry/sentry-unreal/pull/982/commits/19da3b2fe436afcdf5305a3790825677b137deea)).

The idea is to store user-defined attachments in an internal vector and remove them one by one using the native `sentry_remove_attachment` whenever `ClearAttachments` is called. This way, we avoid accidentally removing any attachments that were added during initialization.

@jpnurmi wdyt?

jpnurmi: Sounds ok. In principle, `sentry_clear_attachments()` works the same way with Crashpad to avoid removing the internal `__sentry-event` and `__sentry-breadcrumbN` attachments.

jpnurmi: Btw, not saying we can't or shouldn't improve this later on! :slightly_smiling_face: But I think this is fine for now.

tustanivsky: Alright, I believe we're no longer blocked here and can proceed with the merge. Further improvements related to handling attachments may come in subsequent PRs.

PR #986, opened 2025-07-02
chore: Add line endings and BOM checks to CI lint workflow

tustanivsky:
This PR addresses concerns regarding inconsistent line endings and BOM in some source files included in the plugin package. It also adds an automated check for these issues as part of the CI lint workflow (see #970 for additional details).

Related to #969

#skip-changelog


jpnurmi:
Could it be easier done with [.gitattributes](https://git-scm.com/docs/gitattributes)?

```
*.h text eol=lf encoding=UTF-8
*.cpp text eol=lf encoding=UTF-8
```

And then `git add --renormalize .` once for the existing files.

.github/workflows/lint.yml
```diff
@@ -12,5 +12,27 @@ jobs:
...
+      - name: Check Line Endings
+        run: |
+          echo "Checking for consistent line endings (LF) in source files..."          
+          files_with_crlf=$(find plugin-dev/Source/Sentry plugin-dev/Source/SentryEditor -name "*.h" -o -name "*.cpp" | xargs grep -l $'\r' 2>/dev/null || true)
+          if [ -n "$files_with_crlf" ]; then
+            echo "ERROR: Found files with CRLF line endings. Please convert all source files to use LF line endings."
+            echo "$files_with_crlf"
+            exit 1          
+          fi          
+          echo "✓ All source files have consistent LF line endings."
```
vaind: should be doable with clang-format alone
- https://stackoverflow.com/questions/52262866/clang-format-normalized-line-endings

plugin-dev/Source/Sentry/Private/HAL/PlatformSentryAttachment.h
```diff
@@ -1,21 +1,21 @@
...
+#include "Android/AndroidSentryAttachment.h"
+#elif PLATFORM_APPLE
+#include "Apple/AppleSentryAttachment.h"
+#else
+#include "Null/NullSentryAttachment.h"
+#endif
+
+static TSharedPtr<ISentryAttachment> CreateSharedSentryAttachment(const FString& Path, const FString& Filename, const FString& ContentType)
+{
+	return MakeShareable(new FPlatformSentryAttachment(Path, Filename, ContentType));
```
vaind, line 17: also, any reason to use tabs instead of spaces?

tustanivsky: That's a [coding standard](https://dev.epicgames.com/documentation/en-us/unreal-engine/epic-cplusplus-coding-standard-for-unreal-engine#tabsandindenting) for Unreal.

.github/workflows/lint.yml
tustanivsky: Yes, that's a viable option. However, a separate check here seems to provide [cleaner output](https://github.com/getsentry/sentry-unreal/actions/runs/16020537333/job/45196212853#step:3:13) compared to `clang-format` which tends to flood the logs with `error: code should be clang-formatted` messages for every line in a file with invalid line endings.

plugin-dev/Source/Sentry/Private/HAL/PlatformSentryAttachment.h
vaind: 😿

.github/workflows/lint.yml
vaind: you could just push the clang-formatted code automatically

tustanivsky: Done - `clang-format` now also checks for line endings and modifies sources in-place when needed. If any changes were made they will be pushed in the next step.

.gitattributes
```diff
@@ -0,0 +1,2 @@
+*.h text eol=lf encoding=UTF-8
+*.cpp text eol=lf encoding=UTF-8
```
vaind: nit: any chance you could set up your IDE to add new line at EOF?

.github/workflows/lint.yml
```diff
@@ -9,8 +9,38 @@ jobs:
...
+            echo "FORMATTING_CHANGED=false" >> $GITHUB_OUTPUT
+          else
+            echo "Code formatting changes detected."
+            echo "FORMATTING_CHANGED=true" >> $GITHUB_OUTPUT
+          fi
+
+      - name: Commit Formatted Code
+        if: steps.clang-format.outputs.FORMATTING_CHANGED == 'true'
+        env:
+          GITHUB_BRANCH: ${{ github.ref_name }}
```
vaind: This shouldn't run on the main branch

tustanivsky: Good point - added a corresponding check to `if` block here.

Issue #942, opened 2025-06-02
Investigate how to enable callbacks invocation during garbage collection


tustanivsky:
The way we currently handle `beforeSend` callback (and a few others) requires us to instantiate UObject-derived classes which could be unsafe during GC (related issue for reference - https://github.com/getsentry/sentry-unreal/issues/573). In order to prevent potential issues in this situation we're skipping callback invocation (if one is configured) entirely during GC.

tustanivsky:
What options do we have:

### 1. Don't use UObjects in callback handler interface

We can expose certain internal interfaces (e.g. `ISentryEvent`) to plugin users and use them as callback parameters instead of UObjects. This would mean operating on plain C++ objects which would not be accessible from Blueprints.

An alternative is to switch to using UStructs for the entities passed to the callback handler. However, this would be a significant API change (see the old draft PR for breadcrumbs: https://github.com/getsentry/sentry-unreal/pull/854).

### 2. Pre-allocate UObjects pool

Required UObjects can be created during initialization and reused during garbage collection. Before invoking the callback, the underlying native objects can be assigned to them via some setter function.

### 3. Leave current implementation and document its limitations

### 4. Other considerations

- Waiting for GC to finish can introduce potential timeout or deadlock risks
- Non-fatal events occurring during GC can be queued and re-captured afterward but this does not address the issue for crashes

PR #990, opened 2025-07-09
Improve Unreal SDK callbacks propagation and platform-specific handling

tustanivsky:
This PR improves how Unreal SDK callbacks are handled internally as part of the effort to support their invocation during garbage collection (#942).

Key changes:
- The `beforeSend`, `beforeBreadcrumb` and `traceSampler` callbacks are now propagated to the Unreal layer only if the user has configured their custom handlers in the plugin settings. This allows to avoid instantiating redundant `UObject`'s as platform-specific implementations provide all the required default handling.
- For Android, attaching the game log file was moved to Java leaving the `USentryBeforeSendHandler` empty - it now only serves as a base class for the user's custom implementation.
- For Windows/Linux, `OnCrash` now defers Unreal callback invocation to `OnBeforeSend` while preserving the logic specific for crash events.
- 
#skip-changelog


plugin-dev/Source/Sentry/Private/SentryBeforeSendHandler.cpp
```diff
@@ -2,33 +2,10 @@
...
-#if PLATFORM_ANDROID || PLATFORM_APPLE
-		const FString LogFilePath = Event->IsCrash() ? SentryFileUtils::GetGameLogBackupPath() : SentryFileUtils::GetGameLogPath();
-#else
-		const FString LogFilePath = SentryFileUtils::GetGameLogPath();
-#endif
-
-		Hint->AddAttachment(
-			USentryLibrary::CreateSentryAttachmentWithPath(
-				LogFilePath, FPaths::GetCleanFilename(LogFilePath), TEXT("text/plain")));
-	}
```
jpnurmi, line 31: This logic was moved to the Android subsystem. Just to double-check, was it ever used at all on any other platforms?

jpnurmi (approved):

plugin-dev/Source/Sentry/Private/SentryBeforeSendHandler.cpp
tustanivsky: Not really - platforms other than Android don’t support hints and handle `beforeSend` internally within their respective subsystems. Moving this logic into the Android subsystem brings it in line with the others and more importantly eliminates the need to always instantiate `USentryBeforeSendHandler` during the SDK initialization (later it allows to avoid the unnecessary creation of Unreal objects when the user hasn’t configured any custom handling)

PR #1000, opened 2025-07-14
Fix packaging errors on disabled/unsupported platforms

tustanivsky:
This change resolves build errors that occurred when packaging project for platforms where Sentry is disabled or not supported.

#skip-changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,9 +8,9 @@
 
 ### Dependencies
 
-- Bump CLI from v2.46.0 to v2.47.0 ([#989](https://github.com/getsentry/sentry-unreal/pull/989))
-  - [changelog](https://github.com/getsentry/sentry-cli/blob/master/CHANGELOG.md#2470)
-  - [diff](https://github.com/getsentry/sentry-cli/compare/2.46.0...2.47.0)
+- Bump CLI from v2.46.0 to v2.47.1 ([#989](https://github.com/getsentry/sentry-unreal/pull/989), [#1001](https://github.com/getsentry/sentry-unreal/pull/1001))
+  - [changelog](https://github.com/getsentry/sentry-cli/blob/master/CHANGELOG.md#2471)
+  - [diff](https://github.com/getsentry/sentry-cli/compare/2.46.0...2.47.1)
 - Bump Java SDK (Android) from v8.16.0 to v8.17.0 ([#994](https://github.com/getsentry/sentry-unreal/pull/994))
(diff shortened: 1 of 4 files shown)


plugin-dev/Source/Sentry/Private/HAL/PlatformSentryScope.h
```diff
@@ -6,15 +6,10 @@
...
 #elif USE_SENTRY_NATIVE
 #include "GenericPlatform/GenericPlatformSentryScope.h"
 #else
 #include "Null/NullSentryScope.h"
 #endif
-
-static TSharedPtr<ISentryScope> CreateSharedSentryScope()
-{
-	return MakeShareable(new FPlatformSentryScope);
-}
```
jpnurmi, line 20: out of curiosity, what was the build error?

tustanivsky: Here's an example of error message that showed up while packaging the plugin for Xbox One (currently not supported):
```
UATHelper: Packaging (XB1): D:\projects\sentry-unreal\sample\Plugins\sentry\Source\Sentry\Private\HAL\PlatformSentryScope.h(19,27): error C2061: syntax error: identifier 'FPlatformSentryScope'
UATHelper: Packaging (XB1):    return MakeShareable(new FPlatformSentryScope);
UATHelper: Packaging (XB1):                             ^
```
Technically, Xbox One is `PLATFORM_MICROSOFT` so the `Microsoft/MicrosoftSentryScope.h` is included in this case, however the contents of that header are wrapped in another `USE_SENTRY_NATIVE` check which is only defined in `Sentry.Build.cs` for platforms where `sentry-native` is actually used, e.g.:
https://github.com/getsentry/sentry-unreal/blob/bf91a82fdf91c20ed2afbabcd686b78a892c6d87/plugin-dev/Source/Sentry/Sentry.Build.cs#L131

Since we don't have XB1 extension yet the header ends up being empty meaning we don't have a proper typed for `FPlatformSentryScope` available:
https://github.com/getsentry/sentry-unreal/blob/bf91a82fdf91c20ed2afbabcd686b78a892c6d87/plugin-dev/Source/Sentry/Private/Microsoft/MicrosoftSentryScope.h#L19

jpnurmi: Ok, thanks for the clarification! :pray:

jpnurmi (approved):

Issue #998, opened 2025-07-11
Plugin packaging errors during FAB update submission processing


tustanivsky:
After submitting `1.0.0-beta.4` update to FAB the following errors came up during plugin packaging on Mac:

## UE 5.4
In order to pre-buid the plugin for 5.4 Epic Games uses Xcode 14.2 which misses some Swift compatibility libs that Cocoa SDK depends on.
```
------ Building 28 action(s) started ------
[1/28] (Wall: 5.34s) Compile [Apple] SharedPCH.Core.Cpp20.h
...
[20/28] (Wall: 0.05s) Link [Intel] UnrealEditor-Sentry.dylib_x64
clang: error: no such file or directory: '/Applications/Xcode-v14.2.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift/macosx/libswiftCompatibility56.a'
clang: error: no such file or directory: '/Applications/Xcode-v14.2.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift/macosx/libswiftCompatibilityPacks.a'
[21/28] (Wall: 19.01s) Compile [Apple] Module.Sentry.3.cpp
[22/28] (Wall: 0.05s) Link [Apple] UnrealEditor-Sentry.dylib_arm64
clang: error: no such file or directory: '/Applications/Xcode-v14.2.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift/macosx/libswiftCompatibility56.a'
clang: error: no such file or directory: '/Applications/Xcode-v14.2.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift/macosx/libswiftCompatibilityPacks.a'
...
Total time in Parallel executor: 63.33 seconds
Total execution time: 73.94 seconds
Took 74.15s to run dotnet, ExitCode=6
UnrealBuildTool failed. See log for more details. (/Users/build/Build/U5M-Mac/Sync/Engine/Programs/AutomationTool/Saved/Logs/BuildMarketPlaceItems/UBA-UnrealEditor-Mac-Development.txt)
AutomationTool executed for 0h 1m 15s
AutomationTool exiting with ExitCode=6 (6)
RunUAT ERROR: AutomationTool was unable to run successfully. Exited with code: 6
```
## UE 5.5
Could be related to the Xcode version as well (provided logs don't mention which one was used).
```
[...]

PR #902, opened 2025-05-08
Use `sentry-cocoa` as static lib on Mac

tustanivsky:
This PR changes how Unreal SDK consumes `sentry-cocoa` on Mac.

Switching to a static library allows to address the issue with loading unverified binaries on startup (#476) and simplify calls to native API (`SENTRY_APPLE_CLASS` is no longer required).

#skip-changelog

PR #1012, opened 2025-07-17
Use `sentry-cocoa` as dynamic lib on Mac

tustanivsky:
This effectively reverts the changes made in #902 to resolve issues with publishing the plugin to FAB.

Linking Cocoa as a static library introduces dependencies on Xcode-provided Swift libraries the availability of which can vary between Xcode versions. Since Epic builds the plugin for the three latest Unreal Engine versions (each using a different version of Xcode) it becomes difficult to maintain compatibility across the binaries we ship with the SDK.

Cocoa SDK itself appears to require a newer Xcode making it problematic to pre-build the corresponding library in CI using its older versions (e.g. UE 5.4 needs Xcode 14.2).

Related to #998

#skip-changelog

### DIFF
--- a/plugin-dev/Source/Sentry/Private/SentryModule.cpp
+++ b/plugin-dev/Source/Sentry/Private/SentryModule.cpp
@@ -10,6 +10,7 @@
 #include "Modules/ModuleManager.h"
 #include "UObject/Package.h"
 #include "UObject/UObjectGlobals.h"
+#include "HAL/PlatformProcess.h"
 
 #define LOCTEXT_NAMESPACE "FSentryModule"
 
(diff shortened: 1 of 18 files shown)


plugin-dev/Source/Sentry/Private/SentryModule.cpp
```diff
@@ -23,6 +24,17 @@ void FSentryModule::StartupModule()
 	SentrySettings = NewObject<USentrySettings>(GetTransientPackage(), "SentrySettings", RF_Standalone);
 	SentrySettings->AddToRoot();
 
+#if PLATFORM_MAC
+	// Load Sentry Cocoa dynamic library
+	FString LibraryPath = FPaths::Combine(GetBinariesPath(), TEXT("sentry.dylib"));
+	mDllHandleSentry = FPlatformProcess::GetDllHandle(*LibraryPath);
```
limbonaut, line 30: Are there good reasons not to link at build time?

tustanivsky: Since we only have the dynamic library without the accompanying import library we have to load it and import the required functions manually at runtime.

Additionally, since Unreal doesn't auto-stage dynamic libs it has to be explicitly added to list of runtime dependencies in `Sentry.Build.cs` to put it alongside the executable when packaging the game.

limbonaut (approved):

Issue #1010, opened 2025-07-15
Irrelevant screenshot attached to captured crash event


tustanivsky:
**Steps to reproduce:**
1. Crash app before rendering thread is initialized
2. Screenshot capturing in `HandleOnCrash` fails with the message:
```
Can't take a screenshot when not in the game thread or slate thread
```

**Expected result:**
Crash uploaded without screenshot attachment.

**Actual result:**
Screenshot taken during previous app session is attached to captured crash.

**Platforms:**
Native (Windows/Linux)

**Potential reason:**
Captured screenshots aren’t cleared from disk before `sentry-native` is initialized during the next app startup after a crash.

PR #1019, opened 2025-07-22
Fix irrelevant screenshot being attached to captured crash event

tustanivsky:
This PR fixes an issue where a screenshot from a previous app session could be incorrectly attached to a crash event in the current session. This occurred when screenshot capturing failed during the current session, leaving the old screenshot file in place.

Key changes:
- Screenshot attachments are now added during the `on_crash` hook handling instead of at initialization
- Cached screenshots are now cleared from disk at app startup to prevent stale data

Fixes https://github.com/getsentry/sentry-unreal/issues/1010

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
 - On mobile platforms, the default traces sampler will no longer be created automatically if one is not explicitly configured in the plugin settings (`General -> Performance Monitoring`).
 - If upgrading from a version prior to 0.9.0 legacy settings `DsnUrl`, `EnableVerboseLogging` and `EnableStackTrace` will no longer be read from the project configuration file automatically. Instead, you must re-set them in plugin settings to adopt the new format.
 
+### Fixes
+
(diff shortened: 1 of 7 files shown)


jpnurmi (approved):
Side-note: sentry-native saves screenshots in the unique "run" directory (`<db>/<uuid>/screenshot.png`). That way, screenshots are automatically cleaned up when old runs are either handled or pruned.  Does sentry-unreal have any way to determine that directory? Does it know or have access to the run UUID?

tustanivsky:
> Side-note: sentry-native saves screenshots in the unique "run" directory (`<db>/<uuid>/screenshot.png`). That way, screenshots are automatically cleaned up when old runs are either handled or pruned. Does sentry-unreal have any way to determine that directory? Does it know or have access to the run UUID?

Yes, I've noticed that `sentry-native` implicitly copies attachments into "run" dir when using `crashpad` backend but that doesn't seem to be the case with `breakpad` (currently used in FAB version of the package on Windows and on Xbox). So to support both scenarios screenshots are currently saved to a separate dir which is then cleared on every startup if needed.

PR #1024, opened 2025-07-23
Fix issues in FAB plugin package

tustanivsky:
This PR addresses issues discovered in FAB version of the plugin (`1.0.0-beta.5`) which for the first time includes executables and scripts.

Key changes:
- Added manual copying of `crashpad_handler.exe` to plugin's build output dir in post-build step (by default Epic obfuscates any extra binaries from there before publishing the package)
- Added execute permission to Sentry CLI on Mac in pre-build step (by default Epic set all permissions to read-only)
- Added heuristics to check package type (`github` vs `marketplace`)

#skip-changelog

### DIFF
--- a/plugin-dev/Source/Sentry/Private/SentryModule.cpp
+++ b/plugin-dev/Source/Sentry/Private/SentryModule.cpp
@@ -16,8 +16,6 @@
 
 const FName FSentryModule::ModuleName = "Sentry";
 
-const bool FSentryModule::IsMarketplace = false;
-
 void FSentryModule::StartupModule()
 {
 	// This code will execute after your module is loaded into memory; the exact timing is specified in the .uplugin file per-module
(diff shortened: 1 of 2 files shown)


plugin-dev/Source/Sentry/Private/SentryModule.cpp
```diff
@@ -117,7 +117,16 @@ FString FSentryModule::GetPluginVersion()
 
 bool FSentryModule::IsMarketplaceVersion()
 {
-	return IsMarketplace;
```
jpnurmi, line 120: was `IsMarketplace` left unused? should it be removed?

tustanivsky: Yes, I believe it can be safely removed.

Additional context: `IsMarketplace` value was set by CI script depending on plugin package type (see https://github.com/getsentry/sentry-unreal/pull/1024/commits/5ef08060daba482dc2cf1f62a1fcd210f294af76)

jpnurmi (approved):

PR #1052, opened 2025-08-14
Add support for unreal's additional crash context scopes

redxdev:
Requires Epic to accept https://github.com/EpicGames/UnrealEngine/pull/13686 first as one required method is not exported from `Core`.

Unreal has a little-known macro `UE_ADD_CRASH_CONTEXT_SCOPE` which allows running code to add additional information to a crash if the crash happens within the scope the macro is in. This adds support to sentry to retrieve this information and add it as an additional scope. Some other systems may also add data into this via `OnAdditionalCrashContextDelegate()`.

### DIFF
--- a/plugin-dev/Source/Sentry/Private/GenericPlatform/CrashReporter/GenericPlatformSentryCrashContext.cpp
+++ b/plugin-dev/Source/Sentry/Private/GenericPlatform/CrashReporter/GenericPlatformSentryCrashContext.cpp
@@ -9,7 +9,7 @@
 
 #include "GenericPlatform/GenericPlatformSentryScope.h"
 
-#if WITH_ADDITIONAL_CRASH_CONTEXTS
+#if UE_VERSION_NEWER_THAN(5, 7, 0) && WITH_ADDITIONAL_CRASH_CONTEXTS
 struct FSentryCrashContextExtendedWriter : public FCrashContextExtendedWriter
 {
 public:


plugin-dev/Source/Sentry/Private/GenericPlatform/CrashReporter/GenericPlatformSentryCrashContext.cpp
```diff
@@ -9,6 +9,52 @@
 
 #include "GenericPlatform/GenericPlatformSentryScope.h"
 
+#if WITH_ADDITIONAL_CRASH_CONTEXTS
```
redxdev: I'll add that this block will probably need an `&& UE_VERSION_5_7_OR_GREATER` or whatever the macro is, which I'll add if/when Epic accepts their side of the PR.

redxdev:
Looks like epic grabbed the required commit on their side: https://github.com/EpicGames/UnrealEngine/commit/8d79c7e26943f5c3c3ad1291db25488215fa8b09

I've updated the review to check for a minimum version of 5.7.

redxdev:
We've noticed an issue where this can get into an infinite loop specifically when sending an event that is _not_ a crash (seems to be an issue with some graphics code applying a context scope that probably doesn't expect to be called into when not in an actual crashing state), so this shouldn't be committed as-is - I need to move the application of the engine's context to a location only used when sending an actual crash.

redxdev:
Hm, unfortunately we've now seen the hang caused by the breadcrumbs context even when crashing. So I guess this isn't safe after all.

Issue #910, opened 2025-05-12
Allow players to opt-out of crash reporting


j0tt:
I know sentry native supports requiring consent before sending crash data, because I wrote it into the Unreal Plugin here: https://github.com/getsentry/sentry-unreal/compare/main...j0tt:sentry-unreal:jott/main/toggleconsent?expand=1

I don't, however, know how to implement such a feature on the other platforms (including a "special one" not in the linked compare).

Please implement this feature for all platforms! Consider using the code linked above as a starting point, or do it completely differently, I just figured I'd share what I have in case it's helpful.

PR #1053, opened 2025-08-19
Add functionality to give/revoke user consent for crash uploads

nwhite-riot:
Addresses #910. Huge thanks to @j0tt for putting this code together - I'm merely trying to push it through 😄


plugin-dev/Source/Sentry/Private/SentrySubsystem.cpp
```diff
@@ -466,6 +466,38 @@ void USentrySubsystem::EndSession()
 	SubsystemNativeImpl->EndSession();
 }
 
+void USentrySubsystem::GiveUserConsent()
+{
+	check(SubsystemNativeImpl);
+
+	if (SubsystemNativeImpl && SubsystemNativeImpl->IsEnabled())
```
j0tt:
```suggestion
	if (!SubsystemNativeImpl || SubsystemNativeImpl->IsEnabled())
```

To be consistent with the rest of the file.

nwhite-riot: If Sentry maintainers want me to adhere to this style, I'd be happy to. I'd argue we should invert the other instances because we want the `if` branch to be evaluated on the most likely flow, instead of an awkward early-out

tustanivsky: Let's keep these checks consistent with rest of the file for now. If we decide to change the current style it'd probably make sense to tackle this in a separate PR.

nwhite-riot: Updated!

tustanivsky (approved):

PR #1063, opened 2025-09-02
Add missing `SWIFT_PACKAGE` define for Apple platforms

tustanivsky:
This PR resolves packaging errors in UE 5.4 and 5.5 due to a missing `SWIFT_PACKAGE` define when targeting Apple platforms. The problem was discovered after submitting version `1.0.0` to FAB.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+- Packaging errors in Unreal Engine 5.4 and 5.5 caused by a missing `SWIFT_PACKAGE` define when targeting Mac and iOS
+
 ## 1.0.0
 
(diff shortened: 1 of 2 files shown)


tustanivsky:
> What's the error, out of curiosity? Shouldn't `#if SWIFT_PACKAGE` and `#if !SWIFT_PACKAGE` in sentry-cocoa headers work fine with a C++ compiler without explicitly defining `SWIFT_PACKAGE=0`? Does Unreal packaging process involve reading and parsing those headers with some custom tooling?

```
error: 'SWIFT_PACKAGE' is not defined, evaluates to 0 [-Werror,-Wundef]
```
Unreal builds with `-Wundef` so missing defines trigger warnings which are promoted to errors via `-Werror`. Interestingly, UE 5.6 doesn’t fail on this which might indicate that Epic changed the default warning/error configuration for packaging in that version.

Issue #77, opened 2022-08-22
Integration/smoke test


vaind:
Similarly to the Unity SDK, we should run a (simple) smoke test to check a game builds with the Sentry plugin and errors are captured.
This should ideally happen in CI.

In order to do that for [the currently supported plugin platforms](https://github.com/getsentry/sentry-unreal/blob/main/plugin-dev/Sentry.uplugin#L22), we would need macOS and Windows runners. Unfortunately, it looks like these are not readily available for use in CI and are [supposed to be installed](https://docs.unrealengine.com/4.27/en-US/Basics/InstallingUnrealEngine/) through the EpicGameLauncher in a non-automation-friendly way (a.k.a. clicking).

Therefore, to run this in CI, we have the following options:
* set up standalone macOS and Windows servers and install the appropriate versions manually + take care of server crashes, freezes, updates, etc
* try to avoid EpicGameLauncher, getting hold of the pre-built executable links and installing those in CI
* build the engine from [source](https://github.com/EpicGames/UnrealEngine) in CI - this requires about 90 GiB of free space on Windows, Engine v5.0.3
* build from source and get that to CI while complying with engine licences
* https://github.com/adamrehn/ue4-docker which provides Dockerfiles (we would need to make the built images private to comply with the Unreal EULA) for Linux and Windows.

PR #1064, opened 2025-09-02
chore: Update sample project to enable running integration tests in CI

### LINKED ISSUE #1061: Add a readme describing the example project
We link to https://github.com/getsentry/sentry-unreal/tree/main/sample from Fab

But landing on it leave u to click around to try finding interesting files to learn about the sample
A readme file could give a high level description about what's there. And even link directly to interesting files.

tustanivsky:
This PR adds functionality to sample project's game instance that allows to parse command line input and run test actions accordingly (e.g. trigger crash, capture message).

CI now uploads the sample project build as a GitHub artifact so it can be verified against various environments (e.g., Amazon Linux which is commonly used for hosting game servers). This verification will be automated in subsequent PRs as part of the ongoing effort to improve the existing pipeline.

Related to:
- #77
- #575

Also, it adds `README.md` for sample project and closes #1061.

#skip-changelog


sample/Source/SentryPlayground/SentryPlaygroundGameInstance.cpp
```diff
@@ -2,9 +2,113 @@
...
+		RunCrashTest();
+	}
+	else if (FParse::Param(CommandLine, TEXT("MessageCapture")))
+	{
+		RunMessageTest();
+	}
+
+	// Integration test requested but no action specified
+	CompleteTestWithResult(TEXT("invalid-command"), false, TEXT("Missing command"));
+}
```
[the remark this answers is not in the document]
tustanivsky: Fixed via https://github.com/getsentry/sentry-unreal/pull/1064/commits/02252c43f9f662886ffe3439e7c52f8a9cd195b2

.github/workflows/test-windows.yml
```diff
@@ -79,3 +79,20 @@ jobs:
...
             -NullRHI
+
+      - name: Collect sample test info
+        if: ${{ always() && steps.run-tests.outcome == 'failure' }}
+        uses: actions/upload-artifact@v4
+        with:
+          name: UE ${{ inputs.unreal-version }} sample test report (Windows)
+          path: |
+            C:/workspace/checkout/sample/Saved/Automation
+
```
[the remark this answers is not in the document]
tustanivsky: Fixed via https://github.com/getsentry/sentry-unreal/pull/1064/commits/2bff464321371bdd121f16afa8ac3f5ffaf2d382

sample/Source/SentryPlayground/SentryPlaygroundGameInstance.cpp
```diff
@@ -2,9 +2,115 @@
...
+		TEXT("Context configuration finished"), TEXT("Test"), TEXT("info"), TMap<FString, FSentryVariant>(), ESentryLevel::Info);
+
+	if (FParse::Param(CommandLine, TEXT("CrashCapture")))
+	{
+		RunCrashTest();
+	}
+	else if (FParse::Param(CommandLine, TEXT("MessageCapture")))
+	{
+		RunMessageTest();
+	}
```
vaind: it may simplify using this later in console integration tests if we used same param values: crash-capture, message-capture

sample/Source/SentryPlayground/SentryPlaygroundGameInstance.cpp
```diff
@@ -2,9 +2,115 @@
...
+	else
+	{
+		// Integration test requested but no action specified
+		CompleteTestWithResult(TEXT("invalid-command"), false, TEXT("Missing command"));
+	}
+}
+
+void USentryPlaygroundGameInstance::RunCrashTest()
+{
+	USentryPlaygroundUtils::Terminate(ESentryAppTerminationType::NullPointer);
```
vaind, line 87: For crashes, this enables us to find the crash event in the integration test script that executes the game:

```c
        // Because we don't get the real ID, create a fake one and set it as a tag.
        sentry_uuid_t uuid = sentry_uuid_new_v4();
        char uuid_str[37];
        sentry_uuid_as_string(&uuid, uuid_str);
        LOG("EVENT_CAPTURED: %s\n", uuid_str);
        sentry_set_tag("test.crash_id", uuid_str);
```

sample/Source/SentryPlayground/SentryPlaygroundGameInstance.cpp
tustanivsky: Renamed params for consistency with console platforms (https://github.com/getsentry/sentry-unreal/pull/1064/commits/a0d7a8c886ea3a0abdd67e6352309d9966ca5c5f).

sample/Source/SentryPlayground/SentryPlaygroundGameInstance.cpp
tustanivsky: Added fake event id generation for test crashes (https://github.com/getsentry/sentry-unreal/pull/1064/commits/de625955164d95db29736919e6281cec7a014958).

sample/Source/SentryPlayground/SentryPlaygroundGameInstance.cpp
```diff
@@ -17,9 +17,9 @@ void USentryPlaygroundGameInstance::Init()
 
 	const TCHAR* CommandLine = FCommandLine::Get();
 
-	// Check '-SentryIntegrationTest' commandline argument to decide between running integration tests
+	// Check '-integration-test' commandline argument to decide between running integration tests
 	// or launching the sample app with UI for manual testing
-	if (FParse::Param(FCommandLine::Get(), TEXT("SentryIntegrationTest")))
+	if (FParse::Param(FCommandLine::Get(), TEXT("integration-test")))
```
vaind: maybe run this if there's any arg? For normal startup with UI, there would be none, right?

vaind: Actually I see in the README.md that's not true...

so maybe check if the arg is one of the expected ones (crash-capture/message-capture)

Where I'm coming from is trying to avoid adding additional args and differences between the gdx repos when we can avoid this

vaind: IIRC Unity does exactly this - if it matches any of the expected args, it does the thing, otherwise it shows the UI

sample/Source/SentryPlayground/SentryPlaygroundGameInstance.cpp
vaind: We also need to print it so that the caller can grab it from stdout and search for it. Just like in the pasted code

vaind: UE_LOG(LogSentrySample, Log, TEXT("EVENT_CAPTURED: %s\n"), *EventId);

sample/Source/SentryPlayground/SentryPlaygroundGameInstance.cpp
tustanivsky: That makes sense - I've removed separate `integration-test` parameter.

It was originally added to simplify checking whether to run integration tests in case the number of supported actions grew over time but since there are only two (at least for now) it’s fine to move on without it.

sample/Source/SentryPlayground/SentryPlaygroundGameInstance.cpp
```diff
@@ -2,9 +2,127 @@
...
+	USentryUser* User = NewObject<USentryUser>();
+	User->Initialize();
+	User->SetUsername(TEXT("TestUser"));
+	User->SetEmail(TEXT("user-mail@test.abc"));
+	User->SetId(TEXT("12345"));
+
+	SentrySubsystem->SetUser(User);
+
+	SentrySubsystem->SetTag(TEXT("test.suite"), TEXT("integration"));
+}
```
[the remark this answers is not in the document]
tustanivsky: Good catch with string literals. Currently, it should be sufficient to check `SentrySubsystem` once in `RunIntegrationTest` to keep the actual test methods concise.

PR #1066, opened 2025-09-03
Fix `jstring` lifetime issue for log file path on Android

tustanivsky:
This PR fixes a crash on Android caused by returning an invalid `jstring` from the JNI bridge when resolving game log path required to create file attachment.

Previously, the corresponding utility method used `FScopedJavaObject<jstring>` which deletes the local reference as soon as the temporary goes out of scope. This left Java with a dangling `jstring` reference leading to a `SIGSEGV` when accessing the value in beforeSend handler.


jpnurmi (approved):
> Would it make sense to make `FSentryJavaObjectWrapper::GetJString()` safe (if possible) to avoid the same pitfall in the future?

Although, making it safe would imply allocating copies in all those other places where it's not needed... :/

limbonaut (approved):

Issue #992, opened 2025-07-09
Sentry crashing on stack overflow


swbrownuwe:
If a stack overflow occurs in UE and is caught by Sentry, and that overflow left little stack available, Sentry might trigger its own stack overflow while attempting to process it.

I'm not sure exactly why, sentry-native supports setting a stack guarantee and from CMakeLists.txt looks like it should have defaulted to 64k, but from the trace it looks like we had less in practice.  This is on 1.0.0-beta.2 built via scripts/build-win64-crashpad.sh.

Stack trace from windbg below,

<pre>
# Child-SP          RetAddr               Call Site							
00 0000009a`18000ff0 00007fff`caef7069     ntdll!RtlpUnwindPrologue+0x5e4							
01 0000009a`180010a0 00007fff`caef6593     ntdll!RtlpxVirtualUnwind+0x109							
02 0000009a`18001150 00007fff`caea2eda     ntdll!RtlpWalkFrameChain+0x373							
03 0000009a`180017a0 00007fff`caea2e52     ntdll!RtlWalkFrameChain+0x2a							
04 0000009a`180017d0 00007ff6`4a0ee043     ntdll!RtlCaptureStackBackTrace+0x42							
05 0000009a`18001800 00007ff6`49ecf9dd     Application!FWindowsPlatformStackWalk::CaptureStackBackTrace+0xc3 [C:\Application\UE\Engine\Source\Runtime\Core\Private\Windows\WindowsPlatformStackWalk.cpp @ 412]							
06 0000009a`18001c80 00007ff6`4a105e6d     Application!FGenericPlatformStackWalk::StackWalkAndDump+0x3d [C:\Application\UE\Engine\Source\Runtime\Core\Private\GenericPlatform\GenericPlatformStackWalk.cpp @ 208]							
07 0000009a`18002c20 00007ff6`4f656c02     Application!FWindowsPlatformStackWalk::StackWalkAndDump+0x4d [C:\Application\UE\Engine\Source\Runtime\Core\Private\Windows\WindowsPlatformStackWalk.cpp @ 166]							
08 0000009a`18002c50 00007ff6`4f656ad8     Application!PrintCrashLog+0x72 [C:\Application\UE\Application\Plugins\Sentry\Source\Sentry\Private\Windows\WindowsSentrySubsystem.cpp @ 32]							
[...]

Issue #991, opened 2025-07-09
Sentry crashing when UE's allocator crashes


swbrownuwe:
A somewhat common error in a UE game is memory corruption that leads to UE's allocator crashing.  Sentry catches this error, but in attempting to log it, uses UE's allocator and crashes.  This leads to no Sentry report being generated.

For example, on Windows with crashpad, sentry-native calling into sentry-unreal's PrintVerboseLog() can crash attempting to allocate an FString, the relevant part of a stack trace is:

- FString::FString
- PrintVerboseLog
- sentry__logger_log
- sentry__backend_new
- crashpad::CrashpadClient::StartHandler
- UnhandledExceptionFilter

I last tested this on 1.0.0-beta.2, but it doesn't look like it's been changed so is probably still an issue.  Sentry should be trying to avoid touching UE internals while handling the crash as UE is in a bad state.

swbrownuwe:
The most reliable way to simulate it is to modify UE's allocator to crash on demand.  Here's modifying their malloc to crash if a flag is set (I think the link should be viewable with access to UE's repo):

https://github.com/swbrownuwe/UnrealEngine/commit/71cd7c56b5023f8a7eda887809e54e118d8ce22a

And adding a console command to Sentry, "DamageAllocator", that will set the flag and reproduce this crash in Sentry in FString:

https://github.com/swbrownuwe/sentry-unreal/commit/ce3ccb91e26a54c75d5132e48c2ee6e0ece3e561

Note this only modifies the binnedmalloc allocators, and the editor defaults to a different allocator, so run with -binnedmalloc2 to be sure it's in use.

PR #1081, opened 2025-09-19
fix: Prevent usage of internal UE logger during crash handling

mujacica:
* Prevent usage of internal UE logger during crash handling
* Disable console logging for Windows crashes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixes
+
+- Prevent usage of internal UE logger during crash handling ([#1081](https://github.com/getsentry/sentry-unreal/pull/1081))
+
 ### Dependencies
 
 - Bump Android Gradle Plugin from v5.10.0 to v5.11.0 ([#1078](https://github.com/getsentry/sentry-unreal/pull/1078))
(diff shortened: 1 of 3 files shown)


tustanivsky (approved):

nwhite-riot:
@mujacica, @tustanivsky, can I ask what this was for? It actually makes the log file worse because Unreal doesn't flush logs, losing a number of ticks of logs (including the actual tick before the crash). I'd even opened a ticket to replicate the log flush behavior on Xbox because it improved the output: https://github.com/getsentry/sentry-xbox/issues/36.

This ticket doesn't explain *why* this is necessary.

mujacica:
Hey @nwhite-riot, the reason for this is mainly trying to address https://github.com/getsentry/sentry-unreal/issues/992 and https://github.com/getsentry/sentry-unreal/issues/991. So in general, if anything goes wrong with UE Allocator/Stack, calling UE_LOG would cause a secondary crash of Sentry Subsystem, so we would not get the event at all. The solution that we did here was not ideal, but if you have any ideas on how to support Allocator/Overflow problems, while at the same time flushing the logs when the crash happens, I am glad to discuss.

mujacica:
Sadly, it does not. We can pre-allocate memory in our code, but since both [StackWalk](https://github.com/EpicGames/UnrealEngine/blob/6978b63c8951e57d97048d8424a0bebd637dde1d/Engine/Source/Runtime/Core/Private/GenericPlatform/GenericPlatformStackWalk.cpp#L182) and [Logger](https://github.com/EpicGames/UnrealEngine/blob/6978b63c8951e57d97048d8424a0bebd637dde1d/Engine/Source/Runtime/Core/Private/Logging/LogMacros.cpp#L75) itself are allocating (using Unreal allocators) internally, it doesn't solve the problem.

I spent quite some time testing this, and with poisoning memory allocator and/or causing stack overflow, a call to UE_LOG/GLog would cause secondary crashes. But as I said, maybe there is a way around it. One of the things we were discussing was sending these logs via Sentry Logs.

PR #1091, opened 2025-09-25
feat: Implement structured logging for Unreal (all platforms)

mujacica:
- Added support for Structured Logging in Unreal
- Integrated with Java/Cocoa/Native SDK features
- Added support for OnBeforeLog hook

---

> [!NOTE]
> Introduces structured logging across Unreal with new log API, configurable settings/filters, a beforeLog hook, platform-specific implementations (Android/Apple/Native), UE log forwarding, and tests.
> 
> - **Core/API**:
>   - Add `USentryLog`/`ISentryLog` and `USentryBeforeLogHandler` (new `beforeLog` hook).
>   - Extend `ISentrySubsystem` with `AddLog(...)` and `InitWithSettings(..., USentryBeforeLogHandler*, ...)`.
>   - New settings for structured logging: `EnableStructuredLogging`, `StructuredLoggingCategories`, `StructuredLoggingLevels`, `bSendBreadcrumbsWithStructuredLogging`.
> - **Output/Forwarding**:
>   - `FSentryOutputDevice` forwards UE logs to structured logging based on level/category filters; optional breadcrumbs.
> - **Android**:
>   - `SentryBridgeJava`: enable logs, `beforeLog` callback, level-specific `addLog*` methods.
>   - Converters add `SentryLogLevel` mapping; new `FAndroidSentryLog`.
>   - Subsystem passes `enableStructuredLogging`/`beforeLogHandler` and implements `AddLog`.
> - **Apple (iOS/macOS)**:
>   - Enable logs in options; implement `beforeSendLog` handler.
>   - New `FAppleSentryLog`; subsystem `AddLog` using SDK logger.
> - **Native (Win/Linux)**:
>   - Integrate `sentry_options_set_before_send_log`, enable logs, and map `AddLog` to `sentry_log_*`.
>   - Add `FGenericPlatformSentryLog` and callback safety helper.
> - **Tests**:
>   - Add specs for `SentryBeforeLogHandler` and `SentryLog`.
> - **Packaging**:
>   - Update snapshot; `pack.ps1` adds `BeforeLogHandler` TObjectPtr rewrite.
> - **Docs/Meta**:
>   - `CHANGELOG.md`: add structured logging feature entry.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0de6f0696770add71145988072f913292948533d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


plugin-dev/Source/Sentry/Private/Android/Java/SentryBridgeJava.java
```diff
@@ -37,6 +37,7 @@ public class SentryBridgeJava {
 	public static native void onConfigureScope(long callbackAddr, IScope scope);
 	public static native SentryEvent onBeforeSend(long handlerAddr, SentryEvent event, Hint hint);
 	public static native Breadcrumb onBeforeBreadcrumb(long handlerAddr, Breadcrumb breadcrumb, Hint hint);
+	public static native SentryLogEvent onBeforeLog(long handlerAddr, SentryLogEvent logEvent, Hint hint);
```
tustanivsky: It looks that the C++ implementation for Android `onBeforeLog` is currently missing (see [AndroidSentryJni.cpp](https://github.com/getsentry/sentry-unreal/blob/2ec63a603233f08cc52e61ba5b885cd68751f79b/plugin-dev/Source/Sentry/Private/Android/Jni/AndroidSentryJni.cpp))

plugin-dev/Source/Sentry/Private/SentryLogData.cpp
```diff
@@ -0,0 +1,29 @@
...
+
+void USentryLogData::Initialize()
+{
+	Message = FString();
+	Level = ESentryLevel::Info;
+}
+
+void USentryLogData::SetMessage(const FString& InMessage)
+{
+	Message = InMessage;
```
tustanivsky: Should we wire the set/get methods to the underlying `ISentryLog`? Right now, in `onBeforeLog` we can’t actually read or modify `USentryLogData` properties as their values are always empty/default, no?

plugin-dev/Source/Sentry/Private/Android/Java/SentryBridgeJava.java
mujacica: Hehe yes, forgot that one... So many customizations, still pending Mobile testing.

plugin-dev/Source/Sentry/Private/SentryLogData.cpp
mujacica: Oh yes, was a TODO I forgot to finalize.

plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentryLog.cpp
```diff
@@ -0,0 +1,84 @@
...
+}
+
+sentry_value_t FGenericPlatformSentryLog::GetNativeObject()
+{
+	return Log;
+}
+
+void FGenericPlatformSentryLog::SetMessage(const FString& message)
+{
+	sentry_value_set_by_key(Log, "message", sentry_value_new_string(TCHAR_TO_UTF8(*message)));
```
nwhite-riot: Why `TCHAR_TO_UTF8` instead of `TCHAR_TO_ANSI`? Especially because you do `FCStringAnsi::Strcmp` on the key output in `GetLevel()`

plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentrySubsystem.cpp
```diff
@@ -185,6 +198,42 @@ sentry_value_t FGenericPlatformSentrySubsystem::OnBeforeBreadcrumb(sentry_value_
...
+	}
+
+	USentryBeforeLogHandler* Handler = GetBeforeLogHandler();
+	if (!Handler)
+	{
+		// If custom handler isn't set skip further processing
+		return log;
+	}
+
+	if (FUObjectThreadContext::Get().IsRoutingPostLoad)
```
nwhite-riot: This pattern of checking for `IsRoutingPostLoad` and then for GC is now used multiple times - can we extract the logic in a way that reduces the duplication?

plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentryLog.cpp
nwhite-riot: In fact we should probably replace it with `StringCast<ANSICHAR>` anyway since the macro is deprecated

plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentrySubsystem.h
```diff
@@ -56,6 +57,7 @@ class FGenericPlatformSentrySubsystem : public ISentrySubsystem
 
 	USentryBeforeSendHandler* GetBeforeSendHandler();
 	USentryBeforeBreadcrumbHandler* GetBeforeBreadcrumbHandler();
+	USentryBeforeLogHandler* GetBeforeLogHandler();
```
nwhite-riot:
```suggestion
	USentryBeforeLogHandler* GetBeforeLogHandler() const;
```

And ideally the others would be `const` too, but that's out-of-scope for this change

plugin-dev/Source/Sentry/Private/Microsoft/MicrosoftSentrySubsystem.cpp
```diff
@@ -19,9 +19,9 @@
 #include "GenericPlatform/GenericPlatformMisc.h"
 #endif
 
-void FMicrosoftSentrySubsystem::InitWithSettings(const USentrySettings* Settings, USentryBeforeSendHandler* BeforeSendHandler, USentryBeforeBreadcrumbHandler* BeforeBreadcrumbHandler, USentryTraceSampler* TraceSampler)
+void FMicrosoftSentrySubsystem::InitWithSettings(const USentrySettings* Settings, USentryBeforeSendHandler* BeforeSendHandler, USentryBeforeBreadcrumbHandler* BeforeBreadcrumbHandler, USentryBeforeLogHandler* BeforeLogHandler, USentryTraceSampler* TraceSampler)
```
nwhite-riot, line 22: Is a following change going to be made in https://github.com/getsentry/sentry-xbox?

plugin-dev/Source/Sentry/Private/SentryLogData.cpp
```diff
@@ -0,0 +1,42 @@
+// Copyright (c) 2025 Sentry. All Rights Reserved.
+
+#include "SentryLogData.h"
+
+#include "HAL/PlatformSentryLog.h"
+
+void USentryLogData::Initialize()
```
nwhite-riot: Why is this `USentryLogData` instead of `USentryLog` to match the naming conventions elsewhere?

plugin-dev/Source/Sentry/Public/SentryLogData.h
```diff
@@ -0,0 +1,48 @@
...
+
+#include "SentryLogData.generated.h"
+
+class ISentryLog;
+
+/**
+ * Data structure representing a structured log entry for Sentry.
+ */
+UCLASS(BlueprintType, NotBlueprintable, HideDropdown)
+class SENTRY_API USentryLogData : public UObject, public TSentryImplWrapper<ISentryLog, USentryLogData>
```
nwhite-riot: If we want to avoid ambiguity/be more specific, should we call this `USentryStructuredLog` (and propagate the name elsewhere)?

scripts/packaging/pack.ps1
```diff
@@ -57,6 +57,7 @@ function packFiles()
             $newSentrySubsystemHeader = $newSentrySubsystemHeader `
                 -replace 'USentryBeforeSendHandler\* BeforeSendHandler;', 'TObjectPtr<USentryBeforeSendHandler> BeforeSendHandler;' `
                 -replace 'USentryBeforeBreadcrumbHandler\* BeforeBreadcrumbHandler;', 'TObjectPtr<USentryBeforeBreadcrumbHandler> BeforeBreadcrumbHandler;' `
+                -replace 'USentryBeforeLogHandler\* BeforeLogHandler;', 'TObjectPtr<USentryBeforeLogHandler> BeforeLogHandler;' `
```
nwhite-riot, line 60: Commenting here but perhaps this is for another review - seems like we should go the other way and have the code use `TObjectPtr`, and replace with `UObject*` if the engine version is older than 5.x

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # Changelog
 
-## Unreleased
+### Features
+
+- Implement structured logging for Unreal (all platforms) ([#1091](https://github.com/getsentry/sentry-unreal/pull/1091))
```
nwhite-riot, line 5: When we say "all platforms", are we including console?

mujacica:
Hey @nwhite-riot, thanks for taking a look! I will address your code comments in the next rev, but on a more general note: yes we want this enabled on consoles too, but what was meant with “all platforms” is all platforms in this repo. As soon as this PR is done, I am going to follow up with all console plugin integrations in other repos (breaking changes for init).

Regarding naming: as you might see I was trying different things out, but sticking with one name makes more sense. I will think about it a little more.

plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentryLog.cpp
mujacica: Looking into a little bit more detail here, it seems that we are doing the same thing for breadcrumbs. I will leave it as it is, and create a follow-up to rework all of the internal classes to new StringCast.

plugin-dev/Source/Sentry/Private/Microsoft/MicrosoftSentrySubsystem.cpp
mujacica: Answered in the general comment below, but yes, it will.

CHANGELOG.md
mujacica: Answered in the comment below.

scripts/packaging/pack.ps1
mujacica: @tustanivsky -> for the follow-up

plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentrySubsystem.cpp
mujacica: Created a common function to resolve this, thanks for the comment!

plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentryLog.cpp
```diff
@@ -0,0 +1,89 @@
...
```
tustanivsky:
```suggestion
	sentry_value_set_by_key(Log, "body", sentry_value_new_string(TCHAR_TO_UTF8(*message)));
```
Native SDK [uses](https://github.com/getsentry/sentry-native/blob/695f4a4dd21b27818ab5b0bbfcbbd9affea52f3b/src/sentry_logs.c#L605) `body` key for the actual log message. We should do the same so that it can be read and written in the `beforeLog` handler on Win, Linux and consoles.

plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentryLog.cpp
```diff
@@ -0,0 +1,89 @@
...
+}
+
+FString FGenericPlatformSentryLog::GetMessage() const
+{
+	sentry_value_t message = sentry_value_get_by_key(Log, "message");
```
tustanivsky:
```suggestion
	sentry_value_t message = sentry_value_get_by_key(Log, "body");
```
Same as above.

plugin-dev/Source/Sentry/Private/Android/AndroidSentryLog.cpp
```diff
@@ -0,0 +1,57 @@
...
+	SetupClassMethods();
+
+	SetMessage(message);
+	SetLevel(level);
+}
+
+void FAndroidSentryLog::SetupClassMethods()
+{
+	SetMessageMethod = GetMethod("setMessage", "(Ljava/lang/String;)V");
+	GetMessageMethod = GetMethod("getMessage", "()Ljava/lang/String;");
```
tustanivsky:
```suggestion
	SetMessageMethod = GetMethod("setBody", "(Ljava/lang/String;)V");
	GetMessageMethod = GetMethod("getBody", "()Ljava/lang/String;");
```
This should prevent crashes on Android caused by attempts to call missing Java `setMessage`/`getMessage` methods via JNI.

plugin-dev/Source/Sentry/Private/Android/AndroidSentryLog.cpp
```diff
@@ -0,0 +1,57 @@
...
+	SetLevelMethod = GetMethod("setLevel", "(Lio/sentry/SentryLevel;)V");
+	GetLevelMethod = GetMethod("getLevel", "()Lio/sentry/SentryLevel;");
```
tustanivsky: In Android SDK for log's level there is a separate enum type - `SentryLogLevel`. We should utilize that instead of standard `SentryLevel` in order to call this API.

plugin-dev/Source/Sentry/Public/SentrySubsystem.h
```diff
@@ -97,6 +98,16 @@ class SENTRY_API USentrySubsystem : public UEngineSubsystem
 	void AddBreadcrumbWithParams(const FString& Message, const FString& Category, const FString& Type, const TMap<FString, FSentryVariant>& Data,
 		ESentryLevel Level = ESentryLevel::Info);
 
+	/**
+	 * Add a structured log message to Sentry.
+	 *
+	 * @param Message Log message to add.
+	 * @param Level Log level.
+	 * @param Category Optional category to prepend to the message.
```
tustanivsky: Should we provide a default value for optional parameter?

plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentryLog.cpp
mujacica: Fixing that everywhere right now, just saw that both Android and Apple do the same. We use message for breadcrumbs but body for logs.

plugin-dev/Source/Sentry/Public/SentrySubsystem.h
mujacica: Yes, LogSentrySDK it is!

plugin-dev/Source/Sentry/Private/Apple/AppleSentryLog.cpp
```diff
@@ -0,0 +1,65 @@
...
+FAppleSentryLog::FAppleSentryLog()
+{
+	LogApple = [SENTRY_APPLE_CLASS(SentryLog) alloc];
+	// Initialize required properties
+	LogApple.timestamp = [NSDate date];
+	LogApple.traceId = [[SENTRY_APPLE_CLASS(SentryId) alloc] init];
+	LogApple.body = @"";
+	LogApple.attributes = @{};
+	LogApple.level = SentryStructuredLogLevelDebug;
+}
```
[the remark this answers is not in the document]
mujacica: Intended way for now.

plugin-dev/Source/Sentry/Public/SentrySubsystem.h
```diff
@@ -97,6 +98,16 @@ class SENTRY_API USentrySubsystem : public UEngineSubsystem
...
 
+	/**
+	 * Add a structured log message to Sentry.
+	 *
+	 * @param Body Log body to add.
+	 * @param Level Log level.
+	 * @param Category Optional category to prepend to the body.
+	 */
+	UFUNCTION(BlueprintCallable, Category = "Sentry")
+	void AddLog(const FString& Body, ESentryLevel Level, const FString& Category = TEXT("LogSentrySDK"));
```
tustanivsky:
```suggestion
	void AddLog(const FString& Body, ESentryLevel Level, const FString& Category = TEXT("LogSentrySdk"));
```
Let's align default log category name with what we have in [SentryDefines.h](https://github.com/getsentry/sentry-unreal/blob/d50b016987c704d8c10e17bbf0fab84ee2cb355d/plugin-dev/Source/Sentry/Private/SentryDefines.h#L7)

plugin-dev/Source/Sentry/Private/Apple/AppleSentryLog.cpp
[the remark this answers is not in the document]
mujacica: SentryLog doesn't have init function.

plugin-dev/Source/Sentry/Private/Android/AndroidSentrySubsystem.cpp
```diff
@@ -136,6 +141,51 @@ void FAndroidSentrySubsystem::AddBreadcrumbWithParams(const FString& Message, co
...
+	case ESentryLevel::Info:
+		FSentryJavaObjectWrapper::CallStaticMethod<void>(SentryJavaClasses::SentryBridgeJava, "addLogInfo", "(Ljava/lang/String;)V",
+			*FSentryJavaObjectWrapper::GetJString(FormattedMessage));
+		break;
+	case ESentryLevel::Debug:
+		FSentryJavaObjectWrapper::CallStaticMethod<void>(SentryJavaClasses::SentryBridgeJava, "addLogDebug", "(Ljava/lang/String;)V",
+			*FSentryJavaObjectWrapper::GetJString(FormattedMessage));
+		break;
+	}
+}
```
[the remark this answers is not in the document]
mujacica: It covers all available ESentryLevel values

plugin-dev/Source/Sentry/Private/Tests/SentryBeforeLogHandler.spec.cpp
```diff
@@ -0,0 +1,146 @@
...
+			const FString OriginalBody = TEXT("Original message");
+			const FString ModifiedBody = TEXT("Modified message");
+			const ESentryLevel OriginalLevel = ESentryLevel::Info;
+			const ESentryLevel ModifiedLevel = ESentryLevel::Error;
+
+			TestLog->SetBody(OriginalBody);
+			TestLog->SetLevel(OriginalLevel);
+
+			// Create a custom handler that modifies the log
+			USentryBeforeLogHandler* ModifyingHandler = NewObject<USentryBeforeLogHandler>();
```
tustanivsky: Custom `beforeLog` handler was created here but it's never used.

mujacica: Right, will just modify it directly.

plugin-dev/Source/Sentry/Private/Apple/Infrastructure/AppleSentryConverters.cpp
```diff
@@ -35,6 +35,34 @@ SentryLevel FAppleSentryConverters::SentryLevelToNative(ESentryLevel level)
...
+		nativeLevel = SentryStructuredLogLevelWarn;
+		break;
+	case ESentryLevel::Error:
+		nativeLevel = SentryStructuredLogLevelError;
+		break;
+	case ESentryLevel::Fatal:
+		nativeLevel = SentryStructuredLogLevelFatal;
+		break;
+	default:
+		UE_LOG(LogSentrySdk, Warning, TEXT("Unknown sentry level value used. Debug will be returned."));
```
nwhite-riot:
```suggestion
		UE_LOG(LogSentrySdk, Warning, TEXT("Unknown Sentry level value used. Debug will be returned."));
```

plugin-dev/Source/Sentry/Private/Apple/Infrastructure/AppleSentryConverters.cpp
```diff
@@ -155,6 +183,35 @@ ESentryLevel FAppleSentryConverters::SentryLevelToUnreal(SentryLevel level)
...
+		unrealLevel = ESentryLevel::Warning;
+		break;
+	case SentryStructuredLogLevelError:
+		unrealLevel = ESentryLevel::Error;
+		break;
+	case SentryStructuredLogLevelFatal:
+		unrealLevel = ESentryLevel::Fatal;
+		break;
+	default:
+		UE_LOG(LogSentrySdk, Warning, TEXT("Unknown sentry structured log level value used. Debug will be returned."));
```
nwhite-riot:
```suggestion
		UE_LOG(LogSentrySdk, Warning, TEXT("Unknown Sentry structured log level value used. Debug will be returned."));
```

plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentryLog.cpp
```diff
@@ -0,0 +1,95 @@
...
+		break;
+	case ESentryLevel::Info:
+		levelStr = "info";
+		break;
+	case ESentryLevel::Debug:
+		levelStr = "debug";
+		break;
+	default:
+		// Use default "info" level for unhandled cases
+		break;
```
nwhite-riot, line 64:
```suggestion
	case ESentryLevel::Debug:
		levelStr = "debug";
		break;
   	case ESentryLevel::Info:
	default:
		// Use default "info" level for unhandled cases
		break;
```

I don't know if fall throughs are allowed, but since we set `levelStr` to "info", we can group `Info` and `default`

plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentryLog.cpp
```diff
@@ -0,0 +1,95 @@
...
+	}
+	sentry_value_set_by_key(Log, "level", sentry_value_new_string(levelStr));
+}
+
+ESentryLevel FGenericPlatformSentryLog::GetLevel() const
+{
+	sentry_value_t levelValue = sentry_value_get_by_key(Log, "level");
```
nwhite-riot, line 71: Should we check this for value null instead of automatically sending it to `sentry_value_as_string`?

plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentrySubsystem.cpp
```diff
@@ -111,6 +114,33 @@ static void PrintVerboseLog(sentry_level_t level, const char* message, va_list a
...
+{
+	if (FUObjectThreadContext::Get().IsRoutingPostLoad)
+	{
+		UE_LOG(LogSentrySdk, Log, TEXT("Executing `%s` handler is not allowed during object post-loading."), *handlerName);
+		return false;
+	}
+
+	if (IsGarbageCollecting())
+	{
+		UE_LOG(LogSentrySdk, Log, TEXT("Executing `%s` handler is not allowed during garbage collection."), *handlerName);
```
nwhite-riot, line 137: Should we move the comment previously in `OnBeforeSend` into this too?

plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentrySubsystem.cpp
```diff
@@ -111,6 +114,33 @@ static void PrintVerboseLog(sentry_level_t level, const char* message, va_list a
...
+		return false;
+	}
```
nwhite-riot, line 140:
```suggestion
	if (IsGarbageCollecting())
	{
	    // If event is captured during garbage collection we can't instantiate UObjects safely or obtain a GC lock
		// since it will cause a deadlock (see https://github.com/getsentry/sentry-unreal/issues/850).
		// In this case event will be reported without calling a `beforeSend` handler.
		UE_LOG(LogSentrySdk, Log, TEXT("Executing `%s` handler is not allowed during garbage collection."), *handlerName);
		return false;
	}
```

plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentrySubsystem.cpp
```diff
@@ -125,18 +155,8 @@ sentry_value_t FGenericPlatformSentrySubsystem::OnBeforeSend(sentry_value_t even
 		return event;
 	}
 
-	if (FUObjectThreadContext::Get().IsRoutingPostLoad)
+	if (!IsCallbackSafeToRun(TEXT("beforeSend")))
```
nwhite-riot: Perhaps we can use `__FUNCTION__` here instead?

```cplusplus
#define IS_CALLBACK_SAFE IsCallbackSafeToRun(__FUNCTION__)
...
if (!IS_CALLBACK_SAFE)
{
    // ...
}
```

plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentrySubsystem.cpp
```diff
@@ -361,10 +391,12 @@ void FGenericPlatformSentrySubsystem::InitWithSettings(const USentrySettings* se
 	sentry_options_set_sample_rate(options, settings->SampleRate);
 	sentry_options_set_max_breadcrumbs(options, settings->MaxBreadcrumbs);
 	sentry_options_set_before_send(options, HandleBeforeSend, this);
+	sentry_options_set_before_send_log(options, HandleBeforeLog, this);
```
nwhite-riot, line 395: Should we only set this if `settings->EnableStructuredLogging` is `true`?

plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentrySubsystem.cpp
```diff
@@ -466,6 +498,49 @@ void FGenericPlatformSentrySubsystem::AddBreadcrumbWithParams(const FString& Mes
...
+	else
+	{
+		FormattedMessage = Body;
+	}
+
+	auto MessageCStrConverter = StringCast<ANSICHAR>(*FormattedMessage);
+	const char* MessageCStr = MessageCStrConverter.Get();
+
+	// Use level-specific sentry logging functions
+	switch (Level)
```
nwhite-riot, line 525: Do we need a `default` case?

plugin-dev/Source/Sentry/Private/Tests/SentryBeforeLogHandler.spec.cpp
```diff
@@ -0,0 +1,134 @@
...
+
+		It("should allow modifying log data", [this]()
+		{
+			const FString OriginalBody = TEXT("Original message");
+			const FString ModifiedBody = TEXT("Modified message");
+			const ESentryLevel OriginalLevel = ESentryLevel::Info;
+			const ESentryLevel ModifiedLevel = ESentryLevel::Error;
+
+			TestLog->SetBody(OriginalBody);
+			TestLog->SetLevel(OriginalLevel);
```
nwhite-riot: Should we test body and level here?

plugin-dev/Source/Sentry/Private/Tests/SentryBeforeLogHandler.spec.cpp
```diff
@@ -0,0 +1,134 @@
...
+			TestTrue("BeforeLogHandler should be called", bHandlerCalled);
+
+			UTestBeforeLogHandler::OnTestBeforeLogHandler.Unbind();
+			SentrySubsystem->Close();
+		});
+
+		It("should handle null log data gracefully", [this]()
+		{
+			USentryLog* Result = BeforeLogHandler->HandleBeforeLog(nullptr);
+			TestEqual("Should return nullptr for nullptr input", Result, static_cast<USentryLog*>(nullptr));
```
nwhite-riot, line 83: Why `static_cast` on a `nullptr`?

plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentrySubsystem.cpp
mujacica: Doesn't matter, it's not going to get used even if set.

plugin-dev/Source/Sentry/Private/Tests/SentryLog.spec.cpp
```diff
@@ -0,0 +1,89 @@
...
+			const ESentryLevel TestLevel = ESentryLevel::Warning;
+
+			SentryLog->SetBody(TestBody);
+			SentryLog->SetLevel(TestLevel);
+
+			TestEqual("Log body", SentryLog->GetBody(), TestBody);
+			TestEqual("Log level", SentryLog->GetLevel(), TestLevel);
+		});
+
+		It("should handle different log levels", [this]()
```
nwhite-riot, line 37: Can you add a test for an out-of-bounds `ESentryLevel` value?

plugin-dev/Source/Sentry/Private/Tests/SentryLog.spec.cpp
```diff
@@ -0,0 +1,89 @@
...
+			// Test Error level
+			SentryLog->SetLevel(ESentryLevel::Error);
+			TestEqual("Error level", SentryLog->GetLevel(), ESentryLevel::Error);
+
+			// Test Fatal level
+			SentryLog->SetLevel(ESentryLevel::Fatal);
+			TestEqual("Fatal level", SentryLog->GetLevel(), ESentryLevel::Fatal);
+		});
+
+		It("should handle empty body", [this]()
```
nwhite-riot, line 67: Is it acceptable for a body to be empty space? Something like `TEXT("   ")`?

plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentrySubsystem.cpp
mujacica: We could add it, but not sure we want to do it. ESentryLevel is tightly controlled, so having anything else here would not make a lot of sense. If we change the ESentryLevel (to add/remove levels), deeper code changes would be required.

plugin-dev/Source/Sentry/Private/Tests/SentryLog.spec.cpp
```diff
@@ -0,0 +1,89 @@
...
+			const FString EmptyBody = FString(TEXT(""));
+
+			SentryLog->SetBody(EmptyBody);
+			SentryLog->SetLevel(ESentryLevel::Info);
+
+			TestEqual("Empty body", SentryLog->GetBody(), EmptyBody);
+			TestEqual("Level with empty body", SentryLog->GetLevel(), ESentryLevel::Info);
+		});
+
+		It("should handle long body text", [this]()
```
nwhite-riot: This seems like a bit of an arbitrary test. Are we validating some max size use-case? Can structured logs accept a max size string? The `FString` here is not what I would consider "long", so perhaps it would be better to be more explicit about what we're testing and why?

plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentryLog.cpp
mujacica: We could, however through the whole codebase I tried having all of the levels explicitly check with appropriate level strings attached. Will leave it for now, but at some point it would make sense to rework these Levels anyhow (either go full-Sentry or full-UE).

plugin-dev/Source/Sentry/Public/SentrySettings.h
```diff
@@ -229,6 +258,22 @@ class SENTRY_API USentrySettings : public UObject
 		Meta = (DisplayName = "Max attachment size in bytes", Tooltip = "Max attachment size for each attachment in bytes. Default is 20 MiB compressed but this size is planned to be increased. Please also check the maximum attachment size of Relay to make sure your attachments don't get discarded there: https://docs.sentry.io/product/relay/options/"))
 	int32 MaxAttachmentSize;
 
+	UPROPERTY(Config, EditAnywhere, Category = "General|Structured Logging",
+		Meta = (DisplayName = "Enable structured logging", ToolTip = "Flag indicating whether to enable structured logging that forwards UE_LOG calls to Sentry logger."))
+	bool EnableStructuredLogging;
```
nwhite-riot, line 263: Should this be `bEnableStructuredLogging`?

plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentryLog.cpp
mujacica: sentry_value_as_string will check "thing" validity, and return nullptr if not valid. Then we check the nullptr few lines below.

plugin-dev/Source/Sentry/Private/Tests/SentryBeforeLogHandler.spec.cpp
mujacica: For type safety:

"If expression is a [null pointer value](https://en.cppreference.com/w/cpp/language/pointer.html#Null_pointers), the result is a null pointer value of type target-type." -> https://en.cppreference.com/w/cpp/language/static_cast.html

plugin-dev/Source/Sentry/Private/Tests/SentryLog.spec.cpp
mujacica: ESentryLevel is an enum, what do you mean exactly?

plugin-dev/Source/Sentry/Private/Tests/SentryBeforeLogHandler.spec.cpp
mujacica: Yes, will add

plugin-dev/Source/Sentry/Private/Tests/SentryLog.spec.cpp
mujacica: Yes, any valid string will be accepted (we are not doing any specific handling on the contents of messages). If needed that can be implemented with custom BeforeLog handler.

plugin-dev/Source/Sentry/Private/Tests/SentryLog.spec.cpp
mujacica: Yep, getting rid of it. I intended to do something different there, but testing max envelope sizes is out of scope for this PR.

plugin-dev/Source/Sentry/Public/SentrySettings.h
mujacica: SentrySettings.h is a bit inconsistent there. Most of the Enable* flags don't have "b" in front, that's why no "b" here.

nwhite-riot: FWIW bools prefaced with `b` seems to be an Unreal coding standard. We can leave it for now, but would be good to become consistent with coding standards

plugin-dev/Source/Sentry/Private/Tests/SentryLog.spec.cpp
nwhite-riot: It's possible to create an out-of-bounds enum by taking an integer that's out of range and casting it to an enum, and it's one of the reasons we have `default` cases

mujacica: I mean we can, and indeed I added default and these stress-tests.

plugin-dev/Source/Sentry/Private/Tests/SentryBeforeLogHandler.spec.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+			// Test that we can modify the log data
+			TestLog->SetBody(ModifiedBody);
+			TestLog->SetLevel(ModifiedLevel);
+
+			USentryLog* Result = BeforeLogHandler->HandleBeforeLog(TestLog);
+
+			TestEqual("Modified body should persist", Result->GetBody(), ModifiedBody);
+			TestEqual("Modified level should persist", Result->GetLevel(), ModifiedLevel);
+		});
```
tustanivsky: The test name is somewhat misleading since it doesn’t actually verify log object modification in the `beforeLog` handler. It basically repeats the previous checks twice. I’d suggest either removing it entirely or aligning it with the approach used in the next test.

mujacica: Will do!

PR #1123, opened 2025-10-22
feat: Implement support for SteamOS detection + crashpad support

mujacica:
* Add support for SteamOS
* Detect and correctly set OS
* Enable the SP option for crashpad/native


mujacica:
@limbonaut well I was thinking about it, but not sure I would want "so specific" gaming logic in the native tbh... But we can discuss it.

tustanivsky (approved):

PR #1127, opened 2025-10-27
chore: Add integration tests for Windows and Linux

tustanivsky:
This PR adds integration tests for the Sentry Unreal SDK using PowerShell's Pester framework. The tests validate end-to-end functionality by running the `SentryPlayground` sample application and verifying that events are properly captured and sent to Sentry.

#### Integration Test Framework:

- PowerShell/Pester-based integration tests (`integration-test/`) that execute the pre-built sample app.
- Cover crash capture and message capture scenarios.
- Validate event data structure, user context, tags, breadcrumbs, and stack traces.
- Use CMake to fetch dependencies (e.g., [app-runner](https://github.com/getsentry/app-runner) module for Sentry API integration).
- Includes a comprehensive README with setup and usage instructions.

#### CI/CD Integration:

- New reusable workflows for Linux and Windows integration tests.
- Run integration tests after unit tests across multiple Unreal Engine versions.
- Upload test outputs and logs on failure to assist with debugging.

#### Sample App Enhancements

- Improved log and events flushing to ensure output is captured
- Workarounds for platform-specific issues (log duplication on Linux UE 4.27)

### Important notices:
- On Windows, integration tests run only for UE 5.2 and newer as that’s the first version where disabling the default Crash Reporter became possible. This allows crashes to be captured using the Sentry integration.
- Running the test app pre-built with UE 5.2 or 5.3 on Windows GitHub runners requires the DirectX June 2010 Runtime to be installed (this is also included in the official Epic Games Windows runtime Docker images)

Enabled by:
- #1064
- https://github.com/getsentry/app-runner/pull/12

Related issues:
- #1002 

Closes:
- #77
- #575

#skip-changelog

PR #1146, opened 2025-11-12
chore: Add integration tests for Android

### LINKED ISSUE #964: Add automation to test Android e2e
Similar to the Unity SDK

https://github.com/getsentry/sentry-unity/blob/main/.github/workflows/smoke-test-build-android.yml
https://github.com/getsentry/sentry-unity/blob/main/.github/workflows/smoke-test-run-android.yml

tustanivsky:
This PR adds Android integration tests for the Sentry Unreal SDK using PowerShell's Pester framework. The tests validate end-to-end functionality by running the `SentryPlayground` sample application on mobile devices and verifying that events are properly captured and sent to Sentry.

It includes the necessary CI workflows and test scripts for both local (adb) and cloud (SauceLabs) execution.

The motivation for using [SauceLabs Real Device Cloud](https://saucelabs.com/products/mobile-testing/real-device-cloud) to run tests in CI pipeline is the current limitation of Android emulators which do not support the rendering backends required by Unreal applications.

Depends on:

- https://github.com/getsentry/app-runner/pull/30

### Changes

- Added test script `Integration.Tests.Android.ps1` for running tests locally via `adb` or on SauceLabs devices.
- Updated `ci.yml` to include the new Android test workflow.
- Added `integration-test-android.yml` to run Android integration tests using SauceLabs in CI.
- Expanded `README.md` with comprehensive instructions for running Android tests locally and via SauceLabs, including required environment variables and prerequisites.
- Added `SAUCE_USERNAME` and `SAUCE_ACCESS_KEY` secrets to the repo.

Related to:
- #1127
- #1131

Closes #964

#skip-changelog

### DIFF
--- a/integration-test/Integration.Android.Tests.ps1
+++ b/integration-test/Integration.Android.Tests.ps1
@@ -170,11 +170,11 @@ Describe 'Sentry Unreal Android Integration Tests (<Platform>)' -ForEach $TestTa
         BeforeAll {
             # Crash event is sent during the MESSAGE run (Run 2)
             # But the crash_id comes from the CRASH run (Run 1)
-            $CrashResult = $global:AndroidCrashResult
-            $CrashEvent = $null
+            $script:CrashResult = $global:AndroidCrashResult
+            $script:CrashEvent = $null
     
             # Parse crash event ID from crash run output
-            $eventIds = Get-EventIds -AppOutput $CrashResult.Output -ExpectedCount 1
+            $eventIds = Get-EventIds -AppOutput $script:CrashResult.Output -ExpectedCount 1
     
             if ($eventIds -and $eventIds.Count -gt 0) {
                 Write-Host "Crash ID captured: $($eventIds[0])" -ForegroundColor Cyan
@@ -182,7 +182,7 @@ Describe 'Sentry Unreal Android Integration Tests (<Platform>)' -ForEach $TestTa
     
                 # Fetch crash event using the tag (event was sent during message run)
                 try {
-                    $CrashEvent = Get-SentryTestEvent -TagName 'test.crash_id' -TagValue "$crashId"
+                    $script:CrashEvent = Get-SentryTestEvent -TagName 'test.crash_id' -TagValue "$crashId"
                     Write-Host "Crash event fetched from Sentry successfully" -ForegroundColor Green
                 }
                 catch {
@@ -195,60 +195,60 @@ Describe 'Sentry Unreal Android Integration Tests (<Platform>)' -ForEach $TestTa
         }
     
         It "Should output event ID before crash" {
-            $eventIds = Get-EventIds -AppOutput $CrashResult.Output -ExpectedCount 1
+            $eventIds = Get-EventIds -AppOutput $script:CrashResult.Output -ExpectedCount 1
             $eventIds | Should -Not -BeNullOrEmpty
             $eventIds.Count | Should -Be 1
         }
     
         It "Should capture crash event in Sentry (uploaded during next run)" {
-            $CrashEvent | Should -Not -BeNullOrEmpty
+            $script:CrashEvent | Should -Not -BeNullOrEmpty
         }
     
         It "Should have correct event type and platform" {
-            $CrashEvent.type | Should -Be 'error'
-            $CrashEvent.platform | Should -Be 'native'
+            $script:CrashEvent.type | Should -Be 'error'
+            $script:CrashEvent.platform | Should -Be 'native'
         }
     
         It "Should have exception information" {
-            $CrashEvent.exception | Should -Not -BeNullOrEmpty
-            $CrashEvent.exception.values | Should -Not -BeNullOrEmpty
+            $script:CrashEvent.exception | Should -Not -BeNullOrEmpty
+            $script:CrashEvent.exception.values | Should -Not -BeNullOrEmpty
         }
     
         It "Should have stack trace" {
-            $exception = $CrashEvent.exception.values[0]
+            $exception = $script:CrashEvent.exception.values[0]
             $exception.stacktrace | Should -Not -BeNullOrEmpty
             $exception.stacktrace.frames | Should -Not -BeNullOrEmpty
         }
     
         It "Should have user context" {
-            $CrashEvent.user | Should -Not -BeNullOrEmpty
-            $CrashEvent.user.username | Should -Be 'TestUser'
-            $CrashEvent.user.email | Should -Be 'user-mail@test.abc'
-            $CrashEvent.user.id | Should -Be '12345'
+            $script:CrashEvent.user | Should -Not -BeNullOrEmpty
+            $script:CrashEvent.user.username | Should -Be 'TestUser'
+            $script:CrashEvent.user.email | Should -Be 'user-mail@test.abc'
+            $script:CrashEvent.user.id | Should -Be '12345'
         }
     
         It "Should have test.crash_id tag for correlation" {
-            $tags = $CrashEvent.tags
+            $tags = $script:CrashEvent.tags
             $crashIdTag = $tags | Where-Object { $_.key -eq 'test.crash_id' }
             $crashIdTag | Should -Not -BeNullOrEmpty
             $crashIdTag.value | Should -Not -BeNullOrEmpty
         }
     
         It "Should have integration test tag" {
-            $tags = $CrashEvent.tags
+            $tags = $script:CrashEvent.tags
             ($tags | Where-Object { $_.key -eq 'test.suite' }).value | Should -Be 'integration'
         }
     
         It "Should have breadcrumbs from before crash" {
-            $CrashEvent.breadcrumbs | Should -Not -BeNullOrEmpty
-            $CrashEvent.breadcrumbs.values | Should -Not -BeNullOrEmpty
+            $script:CrashEvent.breadcrumbs | Should -Not -BeNullOrEmpty
+            $script:CrashEvent.breadcrumbs.values | Should -Not -BeNullOrEmpty
         }
     }
 
     Context "Message Capture Tests" {
         BeforeAll {
-            $MessageResult = $global:AndroidMessageResult
-            $MessageEvent = $null
+            $script:MessageResult = $global:AndroidMessageResult
+            $script:MessageEvent = $null
 
             # Parse event ID from output
             $eventIds = Get-EventIds -AppOutput $MessageResult.Output -ExpectedCount 1
@@ -258,7 +258,7 @@ Describe 'Sentry Unreal Android Integration Tests (<Platform>)' -ForEach $TestTa
 
                 # Fetch event from Sentry (with polling)
                 try {
-                    $MessageEvent = Get-SentryTestEvent -EventId $eventIds[0]
+                    $script:MessageEvent = Get-SentryTestEvent -EventId $eventIds[0]
                     Write-Host "Message event fetched from Sentry successfully" -ForegroundColor Green
                 }
                 catch {
@@ -283,32 +283,32 @@ Describe 'Sentry Unreal Android Integration Tests (<Platform>)' -ForEach $TestTa
         }
 
         It "Should capture message event in Sentry" {
-            $MessageEvent | Should -Not -BeNullOrEmpty
+            $script:MessageEvent | Should -Not -BeNullOrEmpty
         }
 
         It "Should have correct platform" {
             # Android events are captured from Java layer, so platform is 'java' not 'native'
-            $MessageEvent.platform | Should -Be 'java'
+            $script:MessageEvent.platform | Should -Be 'java'
         }
 
         It "Should have message content" {
-            $MessageEvent.message | Should -Not -BeNullOrEmpty
-            $MessageEvent.message.formatted | Should -Match 'Integration test message'
+            $script:MessageEvent.message | Should -Not -BeNullOrEmpty
+            $script:MessageEvent.message.formatted | Should -Match 'Integration test message'
         }
 
         It "Should have user context" {
-            $MessageEvent.user | Should -Not -BeNullOrEmpty
-            $MessageEvent.user.username | Should -Be 'TestUser'
+            $script:MessageEvent.user | Should -Not -BeNullOrEmpty
+            $script:MessageEvent.user.username | Should -Be 'TestUser'
         }
 
         It "Should have integration test tag" {
-            $tags = $MessageEvent.tags
+            $tags = $script:MessageEvent.tags
             ($tags | Where-Object { $_.key -eq 'test.suite' }).value | Should -Be 'integration'
         }
 
         It "Should have breadcrumbs" {
-            $MessageEvent.breadcrumbs | Should -Not -BeNullOrEmpty
-            $MessageEvent.breadcrumbs.values | Should -Not -BeNullOrEmpty
+            $script:MessageEvent.breadcrumbs | Should -Not -BeNullOrEmpty
+            $script:MessageEvent.breadcrumbs.values | Should -Not -BeNullOrEmpty
         }
     }
 }
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -205,7 +205,7 @@ jobs:
     secrets: inherit
     strategy:
       fail-fast: false
-      max-parallel: 2
+      max-parallel: 1
       matrix:
         unreal: ['5.4', '5.5', '5.6', '5.7']
     uses: ./.github/workflows/integration-test-android.yml


.github/workflows/integration-test-android.yml
```diff
@@ -0,0 +1,55 @@
...
+
+      - name: Download sample build
+        uses: actions/download-artifact@v4
+        with:
+          name: UE ${{ inputs.unreal-version }} sample build (Android)
+          path: sample-build
+
+      - name: Install Pester
+        shell: pwsh
+        run: Install-Module -Name Pester -Force -SkipPublisherCheck
```
vaind: `Pester: 5.7.1` is already installed in GH, have you had issues without this step?

https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md#powershell-modules

.github/workflows/integration-test-android.yml
```diff
@@ -0,0 +1,55 @@
...
+        env:
+          SENTRY_UNREAL_TEST_DSN: ${{ secrets.SENTRY_UNREAL_TEST_DSN }}
+          SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_API_TOKEN }}
+          SENTRY_UNREAL_TEST_APP_PATH: ${{ github.workspace }}/sample-build/SentryPlayground-arm64.apk
+          UNREAL_VERSION: ${{ inputs.unreal-version }}
+        run: |
+          cd integration-test
+          mkdir build
+          cmake -B build -S .
+          Invoke-Pester Integration.Tests.Android.SauceLabs.ps1 -CI
```
vaind: maybe a bit cleaner (move working dir to param and remove unnecessary mkdir)
```suggestion
      - name: Run integration tests
        id: run-integration-tests
        shell: pwsh
        working-directory: integration-test
        env:
          SENTRY_UNREAL_TEST_DSN: ${{ secrets.SENTRY_UNREAL_TEST_DSN }}
          SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_API_TOKEN }}
          SENTRY_UNREAL_TEST_APP_PATH: ${{ github.workspace }}/sample-build/SentryPlayground-arm64.apk
          UNREAL_VERSION: ${{ inputs.unreal-version }}
        run: |
          cmake -B build -S .
          Invoke-Pester Integration.Tests.Android.SauceLabs.ps1 -CI
```

integration-test/Integration.Android.Adb.Tests.ps1
```diff
@@ -0,0 +1,418 @@
+# Integration tests for Sentry Unreal SDK on Android
```
vaind, line 1: I assume this would be moved to app-runner as a new provider, in which case both android integration test files would be merged to a single one, right?

.github/workflows/integration-test-android.yml
vaind: Also, invoke-pester assumes files are named "*.Tests.ps1" so the filename should be `Android.Adb.Tests.ps1` or similar. In this case it doesn't matter because you're specifying the exact path, but if you wanted to run all integration tests locally, calling `invoke-pester integration-test` would only pick up files ending with ".Tests.ps1"

integration-test/Integration.Android.SauceLabs.Tests.ps1
```diff
@@ -0,0 +1,604 @@
...
+
+    $base64Auth = [Convert]::ToBase64String([Text.Encoding]::ASCII.GetBytes("${username}:${accessKey}"))
+    $headers = @{
+        'Authorization' = "Basic $base64Auth"
+    }
+
+    try {
+        if ($IsMultipart) {
+            # Use curl for multipart uploads (PowerShell's Invoke-WebRequest struggles with this)
+            $curlCmd = "curl -u `"$username`:$accessKey`" -X $Method `"$Uri`" -F `"payload=@$FilePath`" -F `"name=$(Split-Path $FilePath -Leaf)`""
```
vaind: This should ideally use Invoke-Webrequest to be cross-platform. I've noticed the comment but have you tried the Form param? https://stackoverflow.com/a/60323436

integration-test/Integration.Android.SauceLabs.Tests.ps1
```diff
@@ -0,0 +1,604 @@
...
+                Method = $Method
+                Headers = $headers
+            }
+
+            if ($Body) {
+                $params['Body'] = ($Body | ConvertTo-Json -Depth 10)
+                $params['ContentType'] = $ContentType
+            }
+
+            $response = Invoke-RestMethod @params
```
vaind: Invoke-Restmethod should be avoided - it will just return plain string and swallow parsing error. see app-runner which uses invoke-webrequest and convertfrom-json

integration-test/Integration.Android.SauceLabs.Tests.ps1
```diff
@@ -0,0 +1,604 @@
...
+    if (-not (Test-Path $ApkPath)) {
+        throw "APK file not found: $ApkPath"
+    }
+
+    if ($ApkPath -notlike '*.apk') {
+        throw "Package must be an .apk file. Got: $ApkPath"
+    }
+
+    Write-Host "Uploading APK to SauceLabs Storage..." -ForegroundColor Yellow
+    $uploadUri = "https://api.${Region}.saucelabs.com/v1/storage/upload"
```
vaind: Maybe worth checking if the file exists somehow, e.g. by using a hash-based filename. But I don't really know - if it's always rebuilt then not worth the trouble.

sample/Source/SentryPlayground/SentryPlaygroundGameInstance.cpp
```diff
@@ -18,14 +18,16 @@ void USentryPlaygroundGameInstance::Init()
 {
 	Super::Init();
 
-	const TCHAR* CommandLine = FCommandLine::Get();
+	FString CommandLine = FCommandLine::Get();
+
+	UE_LOG(LogSentrySample, Display, TEXT("Startin app with commandline: %s\n"), *CommandLine);
```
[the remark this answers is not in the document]
tustanivsky: Fixed in https://github.com/getsentry/sentry-unreal/pull/1146/commits/f0bb43dc162cd91cc24b290606fe7ddf1de0b00b

.github/workflows/integration-test-android.yml
tustanivsky: Right, this step is redundant - removed via https://github.com/getsentry/sentry-unreal/pull/1146/commits/9eeb76980d6f2373e79405bf8563bc02d318554d

.github/workflows/integration-test-android.yml
tustanivsky: Good point - renamed scripts in https://github.com/getsentry/sentry-unreal/pull/1146/commits/aeb2f78eda2251962983d31355297d6e26f16c55

integration-test/Integration.Android.SauceLabs.Tests.ps1
tustanivsky: CI always produces a fresh .apk so we can leave it as is for now.

integration-test/Integration.Android.Adb.Tests.ps1
tustanivsky: Yes, tests cases are identical for both Adb and SauceLabs so there's no need to keep separate Tests.ps1 scripts for them. It will probably also make sense to introduce a parameter allowing to specify which provider to use.

integration-test/Integration.Android.SauceLabs.Tests.ps1
```diff
@@ -0,0 +1,604 @@
...
+                $logMessages = @($newLogs | ForEach-Object { if ($_) { $_.message } })
+
+                # Check for completion markers
+                if (($logMessages | Where-Object { $_ -match "TEST_RESULT:" }) -or
+                    ($logMessages | Where-Object { $_ -match "Requesting app exit" })) {
+                    Write-Host "Test completion detected in logs" -ForegroundColor Green
+                    $completed = $true
+                    break
+                }
+            }
```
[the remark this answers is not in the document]
tustanivsky: Fixed in https://github.com/getsentry/sentry-unreal/pull/1146/commits/30d3be26f697c44ece979043107dcd9d932c67a7

integration-test/Integration.Android.SauceLabs.Tests.ps1
```diff
@@ -0,0 +1,604 @@
...
+# - SauceLabs account credentials
+# - Environment variables: SENTRY_UNREAL_TEST_DSN, SENTRY_AUTH_TOKEN, SENTRY_UNREAL_TEST_APP_PATH
+#                          SAUCE_USERNAME, SAUCE_ACCESS_KEY, SAUCE_REGION
+
+Set-StrictMode -Version Latest
+$ErrorActionPreference = 'Stop'
+
+# Script-level state for session management
+$script:SessionId = $null
+$script:LogLineCount = 0  # Track log lines read so far (for delta)
```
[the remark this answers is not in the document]
tustanivsky: Fixed in https://github.com/getsentry/sentry-unreal/pull/1146/commits/30d3be26f697c44ece979043107dcd9d932c67a7

integration-test/Integration.Android.SauceLabs.Tests.ps1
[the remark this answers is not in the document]
tustanivsky: Fixed in https://github.com/getsentry/sentry-unreal/pull/1146/commits/d03ecbb5b82c175d193697a864d3011134fbd7e0

integration-test/Integration.Android.SauceLabs.Tests.ps1
```diff
@@ -0,0 +1,604 @@
...
+    $headers = @{
+        'Authorization' = "Basic $base64Auth"
+    }
+
+    try {
+        if ($IsMultipart) {
+            # Use curl for multipart uploads (PowerShell's Invoke-WebRequest struggles with this)
+            $curlCmd = "curl -u `"$username`:$accessKey`" -X $Method `"$Uri`" -F `"payload=@$FilePath`" -F `"name=$(Split-Path $FilePath -Leaf)`""
+            Write-Debug "Executing: $curlCmd"
+            $response = Invoke-Expression $curlCmd | ConvertFrom-Json
```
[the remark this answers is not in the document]
tustanivsky: Fixed in https://github.com/getsentry/sentry-unreal/pull/1146/commits/d03ecbb5b82c175d193697a864d3011134fbd7e0

integration-test/Integration.Android.SauceLabs.Tests.ps1
tustanivsky: Right, now I recall running into similar issue with Unreal tests in sentry-xbox. Fixed it in https://github.com/getsentry/sentry-unreal/pull/1146/commits/d03ecbb5b82c175d193697a864d3011134fbd7e0

integration-test/Integration.Android.SauceLabs.Tests.ps1
tustanivsky: Thanks for the hint - with `-Form` param it works perfectly!

integration-test/Integration.Android.Adb.Tests.ps1
vaind: > It will probably also make sense to introduce a parameter allowing to specify which provider to use.

We can use connect-device already, not sure we need another param:
- `Connect-Device -Platform Adb` (with or without target arg, depends on whether there are multiple devices available locally - we already support default target selection on Switch and PS5)
- `Connect-Device -Platform SauceLabs -Target "Galaxy A15 5G"`

tustanivsky: I mean param for the test script itself so the user can choose which provider to run the tests with.

integration-test/Integration.Android.Adb.Tests.ps1
```diff
@@ -0,0 +1,418 @@
...
+    # Write-Host "Crash test exit code: $($global:AndroidCrashResult.ExitCode)" -ForegroundColor Cyan
+
+    # ==========================================
+    # RUN 2: Message test - uploads crash from Run 1 + captures message
+    # ==========================================
+    # Currently we need to run again so that Sentry sends the crash event captured during the previous app session.
+    # TODO: use -SkipReinstall to preserve the crash state.
+
+    Write-Host "Running message-capture test (will upload crash from previous run)..." -ForegroundColor Yellow
+    $cmdlineMessageArgs = "-e cmdline '-message-capture'"
```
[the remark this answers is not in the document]
tustanivsky: Fixed in https://github.com/getsentry/sentry-unreal/pull/1146/commits/532fa784e207ae9797689cd3b5d58500db4e8c35

integration-test/Integration.Android.Adb.Tests.ps1
vaind: ah right, that makes sense. Could also be automatic by the presence of saucelabs api key (so it picks what's available locally) but I agree it makes sense to also provide a custom param. Alternatively, you can have both as a [matrix](https://pester.dev/docs/usage/data-driven-tests), like [here where we pick available consoles](https://github.com/getsentry/app-runner/blob/main/app-runner/Tests/Device.Tests.ps1#L125) and you can later filter by adding -ExcludeTag or -FilterTag. But in reality it's just about what you want to achieve and how you use it so up to you - just noting some options.

integration-test/Integration.Android.Adb.Tests.ps1
```diff
@@ -0,0 +1,418 @@
...
+
+    # Check for existing installation
+    Write-Debug "Checking for existing package: $PackageName"
+    $installed = adb -s $DeviceId shell pm list packages | Select-String -Pattern $PackageName -SimpleMatch
+
+    if ($installed) {
+        Write-Host "Uninstalling previous version..." -ForegroundColor Yellow
+        adb -s $DeviceId uninstall $PackageName | Out-Null
+        Start-Sleep -Seconds 1
+    }
```
limbonaut: This seems unnecessary, as `adb install -r` is already being used below (`-r` -- reinstall).

integration-test/Integration.Android.Adb.Tests.ps1
```diff
@@ -0,0 +1,418 @@
...
+
+    # Install APK to device
+    Write-Host "Installing APK to Android device..." -ForegroundColor Yellow
+    Install-AndroidApp -ApkPath $script:ApkPath -PackageName $script:PackageName -DeviceId $script:DeviceId
+
+    # ==========================================
+    # RUN 1: Crash test - creates minidump
+    # ==========================================
+    # The crash is captured but NOT uploaded yet (Android behavior).
+    # TODO: Re-enable once Android SDK tag persistence is fixed (`test.crash_id` tag set before crash is not synced to the captured crash on Android)
```
limbonaut: Do you know if there is an issue to track?

integration-test/Integration.Android.Adb.Tests.ps1
```diff
@@ -0,0 +1,418 @@
...
+    # $cmdlineCrashArgs = "-e cmdline -crash-capture"
+    # $global:AndroidCrashResult = Invoke-AndroidApp -ExecutablePath $script:ActivityName -Arguments $cmdlineCrashArgs
+
+    # Write-Host "Crash test exit code: $($global:AndroidCrashResult.ExitCode)" -ForegroundColor Cyan
+
+    # ==========================================
+    # RUN 2: Message test - uploads crash from Run 1 + captures message
+    # ==========================================
+    # Currently we need to run again so that Sentry sends the crash event captured during the previous app session.
+    # TODO: use -SkipReinstall to preserve the crash state.
```
limbonaut: In Godot tests, I've added a "crash-send" command, which basically launches, initializes Sentry, waits like 10 iterations, and then quits. Maybe we could run a similar test action here after each run for relevant platforms like Android and Cocoa. This would help catch crash reports for each run, even if an App crashes on exit, for example.

integration-test/Integration.Android.SauceLabs.Tests.ps1
```diff
@@ -0,0 +1,584 @@
...
+        }
+
+        try {
+            $stateResponse = Invoke-SauceLabsApi -Method POST -Uri "$baseUri/execute/sync" -Body $stateBody
+            $appState = $stateResponse.value
+
+            Write-Debug "App state: $appState (elapsed: $([int]((Get-Date) - $startTime).TotalSeconds)s)"
+
+            # State 1 = not running, 0 = not installed
+            if ($appState -eq 1 -or $appState -eq 0) {
```
limbonaut: Remark: This is not much to work with. I was hoping Sauce Labs would give a bit more information, like if the application actually crashed or exited cleanly.

integration-test/Integration.Android.SauceLabs.Tests.ps1
```diff
@@ -0,0 +1,584 @@
...
+        -StorageId $script:StorageId `
+        -DeviceName $script:DeviceName `
+        -Region $script:SauceRegion `
+        -UnrealVersion $script:UnrealVersion
+
+    # ==========================================
+    # RUN 1: Crash test - creates minidump
+    # ==========================================
+    # The crash is captured but NOT uploaded yet (Android behavior).
+    # TODO: Re-enable once Android SDK tag persistence is fixed (`test.crash_id` tag set before crash is not synced to the captured crash on Android)
```
limbonaut: Do you think we can grab crash ID from sentry-native output in app-runner?

limbonaut (approved):
I can't say I understand the nuances behind Sauce Labs / Appium APIs. The test script looks fine. It's a shame that crash testing doesn't work. I left a few comments that may be helpful.

.github/workflows/integration-test-android.yml
```diff
@@ -0,0 +1,52 @@
...
+        shell: pwsh
+        working-directory: integration-test
+        env:
+          SENTRY_UNREAL_TEST_DSN: ${{ secrets.SENTRY_UNREAL_TEST_DSN }}
+          SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_API_TOKEN }}
+          SENTRY_UNREAL_TEST_APP_PATH: ${{ github.workspace }}/sample-build/SentryPlayground-arm64.apk
+        run: |
+          cmake -B build -S .
+          $Container = New-PesterContainer -Path 'Integration.Android.Tests.ps1' -Data @{ Platform = 'SauceLabs' }
+          Invoke-Pester -Container $Container -CI
```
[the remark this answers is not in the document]
tustanivsky: No need to do that as Pester in pre-installed on GitHub runners.

integration-test/Integration.Android.Adb.Tests.ps1
tustanivsky: These two look relevant:
- https://github.com/getsentry/sentry-dotnet/issues/4324
- https://github.com/getsentry/sentry-java/issues/2056

integration-test/Integration.Android.SauceLabs.Tests.ps1
tustanivsky: I’m not sure the app-runner is the right place for this. Parsing the app’s output seems test-specific and each gaming SDK that will be using app-runner may have its custom logic for test checks.

integration-test/Integration.Android.Adb.Tests.ps1
tustanivsky: Good point, will address this in the app-runner's implementation of Adb provider.

integration-test/Integration.Android.Adb.Tests.ps1
tustanivsky: Fixed via https://github.com/getsentry/sentry-unreal/pull/1146/commits/a6fda8b81d6d912b1c060b3d9c738c6693401b11

integration-test/Integration.Android.Tests.ps1
```diff
@@ -0,0 +1,275 @@
...
+}
+
+Describe "Sentry Unreal Android Integration Tests ($Platform)" {
+
+    Context "Crash Capture Tests" {
+        BeforeAll {
+            # Crash event is sent during the MESSAGE run (Run 2)
+            # But the crash_id comes from the CRASH run (Run 1)
+            $CrashResult = $global:AndroidCrashResult
+            $CrashEvent = $null
```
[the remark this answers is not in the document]
tustanivsky: Fixed via https://github.com/getsentry/sentry-unreal/pull/1146/commits/622670840ed7732033ba3a9af509a831bf2476e0

Issue #899, opened 2025-05-06
Provide configuration to disable instantiation of crash reporter


nwhite-riot:
If a game wants to exclusively use Sentry for crash handling and reporting (i.e. not using Unreal CRC), we should disable instantiation of `FGenericPlatformSentryCrashReporter`.

tustanivsky:
While looking into this I've noticed an issue somewhat related to CRC  which should be addressed in #901

PR #1178, opened 2025-12-22
Add setting allowing to opt-out Crash Reporter wrapper instantiation

tustanivsky:
This PR adds a new plugin setting that allows disabling the automatic propagation of additional data (e.g., tags and contexts) to Unreal Engine’s Crash Reporter. Disabling this is useful when Crash Reporter is not used/needed as it allows to avoid unnecessary instantiation of the `FGenericPlatformSentryCrashContext` wrapper object.

Related to:
- #901 

Closes:
- #899

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 - Add plugin setting to opt out of automatic Crash Reporter context propagation ([#1178](https://github.com/getsentry/sentry-unreal/pull/1178))
 
+### Fixes
+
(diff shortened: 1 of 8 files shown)


limbonaut (approved):
Can it be auto-detected if Unreal crash reporter is used instead of adding a setting?

tustanivsky:
> Can it be auto-detected if Unreal crash reporter is used instead of adding a setting?

The suggested setting only gives the user control over whether a corresponding wrapper object is created for custom context propagation. The Crash Reporter may still be used even when this setting is disabled (e.g. in older Unreal Engine versions where crashpad integration isn't supported) so auto-detection may not always work as expected.

PR #1293, opened 2026-03-17
Add automatic frame time metrics

### LINKED ISSUE #1281: Game Performance Metrics Auto-Instrumentation (Frame Tracking)
Sentry Unreal SDK currently provides manual APIs for performance monitoring (transactions, spans, metrics), but no automatic instrumentation for common performance signals like frame time, FPS, etc. Developers must manually add `AddDistribution` calls in their game loop to get any production performance data.

The goal is to give every Sentry user production performance visibility with zero manual instrumentation - just enable the feature in plugin settings.

#### Frame work time

Frame duration can be measured two ways:

1. Wall-clock duration between `OnBeginFrame` and `OnEndFrame` on the game thread.

2. Engine’s `IPerformanceDataConsumer` interface - registering a consumer via `GEngine->AddPerformanceDataConsumer()` causes the engine to call `ProcessFrame(FFrameData)` every frame with precomputed timing data. `FFrameData.TrueDeltaSeconds` provides the overall frame duration.

Both approaches produce near-identical frame duration values. The wall-clock method is simpler and self-contained. The performance consumer provides richer data but activates the engine's `AnalyzeFrame()` processing every frame (lightweight - reads globals and does  arithmetic, no allocations or I/O).

> An important nuance: `FApp::GetDeltaTime()` is not suitable here because it includes frame rate limiter sleep time. If a game targets 60fps and a frame completes in 8ms, `GetDeltaTime()` reports ~16.67ms (the engine slept for the remainder). This masks actual workload. Measuring the delegate-to-delegate interval gives the **actual work time**, which is what reveals performance problems.

#### FPS

`GAverageFPS` is a shipping-safe global the engine already computes with an exponential moving average. However, FPS averages may hide problems - a game running at 60fps for 59 frames then 16fps for 1 frame averages ~57fps, which looks fine. The frame time distribution captures this through percentiles (p50, p95, p99), which is strictly more informative. It's possible to add this straight away as there's always a demand for a simple FPS number in dashboards.

#### Per-thread frame time distributions

Emitting separate `game.perf.game_thread`, `game.perf.render_thread`, and `game.perf.gpu` distributions would give full visibility into where time is spent each frame.  The engine's `IPerformanceDataConsumer` interface provides exactly this data (`FFrameData` includes per-thread times), and it's available in shipping builds when a consumer is registered.

#### Thread boundedness

Tracking whether each frame is bottlenecked by game thread, render thread, or GPU. This would answer "where should I optimize?" by emitting a per-frame counter tagged with the slowest thread. This data comes from the same `IPerformanceDataConsumer` interface as per-thread times.

#### Render thread frame time

The engine provides `FCoreDelegates::OnBeginFrameRT` / `OnEndFrameRT` which fire on the render thread. It's possible to independently measure render thread frame duration the same way as game thread.

> Total wall time on the render thread between frame boundaries includes idle/wait time if the render thread is waiting on the game thread so measuring delta can be misleading because it inflates when the game thread is the bottleneck, not rendering.

#### Slow frame counting

After measuring frame time, it's compared against a configurable threshold. This provides a simple, alertable signal: "how often are frames exceeding budget?" The default threshold of 33.3ms (30 FPS) was chosen as a floor that most games consider unacceptable, but this needs to be configurable since a VR game at 90fps has an 11.1ms budget while a strategy game may tolerate 33ms.

> After some early testing this metric was dropped for two reasons:
> 
> 1. Redundant with the distribution - The `game.perf.frame_duration` distribution already captures all the necessary information. Sentry dashboards support querying percentiles (p95, p99) and filtering by threshold directly on distributions, making a separate counter unnecessary.
> 
> 2. Loss of context - a counter tells you “42 slow frames happened” but not how severe they were - a `34 ms` frame and a `500 ms` freeze would both count as `+1`. The distribution preserves the actual values, making it possible to distinguish minor budget overruns from major hitches.

### Proposed metrics

| Metric                     | Type         | Unit        | Description                                                                 |
|---------------------------|--------------|-------------|-----------------------------------------------------------------------------|
| game.perf.frame_time  | Distribution | millisecond | Frame time from engine's performance tracking (`FFrameData.TrueDeltaSeconds`) |
| game.perf.fps             | Gauge        | —           | Engine-smoothed average FPS (`GAverageFPS`)                                |
| game.perf.game_thread     | Distribution | millisecond | Game thread work time (`FFrameData.GameThreadTimeSeconds`)                 |
| game.perf.render_thread   | Distribution | millisecond | Render thread work time (`FFrameData.RenderThreadTimeSeconds`)             |
| game.perf.gpu             | Distribution | millisecond | GPU frame time (`FFrameData.GPUTimeSeconds`)                               |

### Sampling

Emitting a metric every frame would flood the native SDK's internal metrics batcher (at 60 FPS the 100-item buffer fills in under 2 seconds, triggering HTTP flushes almost every second). To avoid this, frame-tracking metric emission can be sampled: only every *N*-th frame is emitted, controlled by the configurable `FrameTimeSampleInterval` setting (default: `30`). At 60 FPS this produces ~2 data points per second, which is sufficient for accurate percentile computation while keeping the buffer fill rate well within the 5-second flush interval.

The tradeoff is that frames between samples are invisible - a hitch on a non-sampled frame is missed. However, across many players and sessions, the aggregate distribution converges to the true distribution even with sparse per-client sampling. The accuracy can be improved by decreasing the sample interval at the cost of higher metric throughput.

#### Alternative sampling methods considered

- **Exponential moving average (EMA)** - Maintains a smoothed frame time (75% previous + 25% current, same as `GAverageFPS`). No frames are lost, but emitting EMA as a distribution leads to “average of averages,” making percentiles hard to interpret. Works as a gauge, but we already emit `game.perf.fps` (same signal in different units).

- **Simple average per window** - Aggregates N frames and emits the mean. All frames contribute, but hitches are diluted (e.g., an 80 ms spike becomes barely noticeable). Also suffers from the “average of averages” issue in distributions.

- **Worst frame per window** - Emits the max frame time over N frames. Never misses a hitch, but introduces layered aggregation (e.g., p95 of per-window maxes ≠ p95 of raw frames), which is harder to reason about.

- **Adaptive sampling** - Emits at a fixed interval plus immediately on hitch frames (with cooldown). Captures all hitches accurately, but adds complexity for limited benefit over raw sampling, since distributions converge across many clients.

### Metric attributes

Metrics are aggregated server-side in Sentry, so event contexts (GPU, CPU, device info) don't carry over - only **metric attributes** survive as filterable dimensions. Without hardware attributes on metrics, it's not possible to answer "what's the frame time distribution on RTX 3060 vs GTX 1060?"

The plugin already collects hardware info during init (`AddGpuContext`, `AddDeviceContext`), but only as event context. A subset needs to be attached as metric attributes.

The key constraint is **cardinality** since every unique attribute value combination creates a separate time series. Each available field was evaluated:

| Attribute | Source | Cardinality | Verdict |
|---|---|---|---|
| `gpu.name` | `FPlatformMisc::GetGPUDriverInfo().DeviceDescription` | ~100-200 common GPUs | **Include** - primary perf differentiator |
| `cpu_cores` | `FPlatformMisc::NumberOfCores()` | ~10 values | **Include** - coarse CPU tier |
| `ram_gb` | `FPlatformMemory::GetConstants().TotalPhysicalGB` | ~8 values | **Include** - memory tier |
| `resolution` | `GSystemResolution.ResX` x `ResY` | ~20 common | **Include** - higher res = more GPU load |
| `map` | Current world/map name | May vary | **Include** - identifies problematic levels |
| `cpu_description` (full brand) | `FPlatformMisc::GetCPUBrand()` | Thousands of unique strings | **Exclude** - cardinality too high |
| `gpu.driver_version` | `GpuDriverInfo.UserDriverVersion` | Changes with every driver update | **Exclude** - cardinality too high |
| `gpu.vendor_name` | `GpuDriverInfo.ProviderName` | ~3 values | **Exclude** - redundant, GPU name implies vendor |

Note: `release` (version) and OS context are included in metrics automatically by the SDK, so regression tracking and platform comparison work without extra attributes.

Hardware attributes are static and can be cached once on start. `map` and `resolution` need dynamic updates (on map change / resolution change).

> To reduce the total amount of time series it might be wort to bucket possible hardware combinations into tiers (e.g. low/mid/high/ultra)

### What this enables

- **Production baselines**: p50/p95/p99 frame time percentiles across real player hardware, not just dev-kit benchmarks
- **Regression detection**: comparing frame time distributions between releases - "p95 went from 14ms to 22ms after the content patch"
- **Map-level profiling**: filtering by `map` attribute to identify which levels have performance problems
- **Hardware tier analysis**: slicing by `gpu.name` / `ram_gb` / `resolution` to determine minimum specs and guide quality presets

## Related items

- https://github.com/getsentry/sentry-unreal/issues/13

tustanivsky:
This PR adds automatic collection of FPS, frame time, and per-thread performance metrics using the engine’s `IPerformanceDataConsumer` interface (see [additional details](https://github.com/getsentry/sentry-unreal/issues/1281)).

## Key changes

- **Automatic frame time metrics (experimental):** Hooks into the engine’s shipping-safe `IPerformanceDataConsumer` API to emit precomputed per-thread timing data; emits `game.perf.frame_duration`, `game.perf.game_thread`, `game.perf.render_thread`, `game.perf.gpu` (distributions, ms) and `game.perf.fps` (gauge).
- **Sampling:** Metrics are emitted every *N*-th frame (`FrameTimeSampleInterval`, default: `30`) to avoid flooding the native SDK’s batcher.
- **Hardware and map attributes:** All metrics include `gpu.name`, `cpu.cores`, `ram.gb`, `res.x`, `res.y` (cached at init) and `map` (updated on level load) for dashboard segmentation.
- **New settings:** `EnableAutoFrameTimeMetrics` and `FrameTimeSampleInterval` under `General | Metrics | Experimental`, both gated by `EnableMetrics` and disabled by default.

Closes  #1281

## Related items

- https://github.com/getsentry/sentry-unreal/issues/1281

## Documentation

- https://github.com/getsentry/sentry-docs/pull/16987

### DIFF
--- a/plugin-dev/Source/Sentry/Private/Utils/SentryPerformanceConsumer.cpp
+++ b/plugin-dev/Source/Sentry/Private/Utils/SentryPerformanceConsumer.cpp
@@ -62,7 +62,7 @@ void FSentryPerformanceConsumer::ProcessFrame(const FFrameData& FrameData)
 
 	const FSentryUnit Ms(ESentryUnit::Millisecond);
 
-	Sentry->AddDistributionWithAttributes(TEXT("game.perf.frame_duration"), FrameData.TrueDeltaSeconds * 1000.0f, Ms, MetricAttributes);
+	Sentry->AddDistributionWithAttributes(TEXT("game.perf.frame_time"), FrameData.TrueDeltaSeconds * 1000.0f, Ms, MetricAttributes);
 	Sentry->AddDistributionWithAttributes(TEXT("game.perf.game_thread"), FrameData.GameThreadTimeSeconds * 1000.0f, Ms, MetricAttributes);
 	Sentry->AddDistributionWithAttributes(TEXT("game.perf.render_thread"), FrameData.RenderThreadTimeSeconds * 1000.0f, Ms, MetricAttributes);
 	Sentry->AddDistributionWithAttributes(TEXT("game.perf.gpu"), FrameData.GPUTimeSeconds * 1000.0f, Ms, MetricAttributes);
--- a/plugin-dev/Source/Sentry/Private/Utils/SentryPerformanceConsumer.h
+++ b/plugin-dev/Source/Sentry/Private/Utils/SentryPerformanceConsumer.h
@@ -10,7 +10,7 @@
  * Consumes per-frame performance data from the engine's FPerformanceTrackingSystem.
  *
  * Registered via GEngine->AddPerformanceDataConsumer(). Emits sampled metrics:
- * - game.perf.duration (distribution, milliseconds) — frame time from engine tracking
+ * - game.perf.frame_time (distribution, milliseconds) — frame time from engine tracking
  * - game.perf.fps (gauge) — engine-smoothed average FPS
  * - game.perf.game_thread (distribution, milliseconds) — game thread work time
  * - game.perf.render_thread (distribution, milliseconds) — render thread work time


plugin-dev/Source/Sentry/Private/SentrySubsystem.cpp
```diff
@@ -1126,6 +1143,16 @@ void USentrySubsystem::ConfigureHangTracking()
...
 
+void USentrySubsystem::ConfigurePerformanceConsumer()
+{
+	PerformanceConsumer = MakeShared<FSentryPerformanceConsumer>();
+
+	if (GEngine)
+	{
+		GEngine->AddPerformanceDataConsumer(PerformanceConsumer);
+	}
+}
```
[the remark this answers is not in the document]
tustanivsky: Not a real issue. `USentrySubsystem` is a `UEngineSubsystem`, meaning it is created and initialized by `GEngine` during engine startup. By the time `Initialize()` runs, `GEngine` is guaranteed to be valid. If `GEngine` were null, the subsystem wouldn’t exist in the first place.

plugin-dev/Source/Sentry/Private/Utils/SentryPerformanceConsumer.cpp
```diff
@@ -0,0 +1,103 @@
...
+		MetricAttributes.Add(TEXT("map"), World->GetMapName());
+	}
+}
+
+void FSentryPerformanceConsumer::CacheAttributes()
+{
+	FGPUDriverInfo GpuDriverInfo = FPlatformMisc::GetGPUDriverInfo(FPlatformMisc::GetPrimaryGPUBrand());
+	if (GpuDriverInfo.IsValid())
+	{
+		MetricAttributes.Add(TEXT("gpu.name"), GpuDriverInfo.DeviceDescription);
```
limbonaut, line 90: Neat!

plugin-dev/Source/Sentry/Private/Utils/SentryPerformanceConsumer.h
```diff
@@ -0,0 +1,42 @@
...
+
+#include "ChartCreation.h"
+#include "CoreMinimal.h"
+#include "SentryVariant.h"
+
+/**
+ * Consumes per-frame performance data from the engine's FPerformanceTrackingSystem.
+ *
+ * Registered via GEngine->AddPerformanceDataConsumer(). Emits sampled metrics:
+ * - game.perf.duration (distribution, milliseconds) — frame time from engine tracking
```
limbonaut: nit: Comment refers to `game.perf.frame_duration`?

limbonaut: I'd call it `frame_time` - that's a more common name for it (unless Unreal uses this specific term).

plugin-dev/Source/Sentry/Private/Utils/SentryPerformanceConsumer.cpp
```diff
@@ -0,0 +1,103 @@
...
+FSentryPerformanceConsumer::FSentryPerformanceConsumer()
+	: SampleInterval(1)
+	, FrameCount(0)
+{
+	const USentrySettings* Settings = FSentryModule::Get().GetSettings();
+	check(Settings);
+
+	SampleInterval = FMath::Max(Settings->FrameTimeSampleInterval, 1);
+
+	CacheAttributes();
```
jpnurmi, line 30: Q: is there any risk these could change during the game lifetime? does it make sense to cache on start or would it add unnecessary overhead?

plugin-dev/Source/Sentry/Private/Utils/SentryPerformanceConsumer.h
tustanivsky: Agreed, let’s use `frame_time`.

plugin-dev/Source/Sentry/Private/Utils/SentryPerformanceConsumer.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+void FSentryPerformanceConsumer::StopCharting()
+{
+}
+
+void FSentryPerformanceConsumer::OnMapLoaded(UWorld* World)
+{
+	if (World)
+	{
+		MetricAttributes.Add(TEXT("map"), World->GetMapName());
```
jpnurmi, line 81: Q: is unsynchronized access safe? are `OnMapLoaded` and `ProcessFrame` called from the same thread, or can loading and rendering happen in parallel?

tustanivsky: Yes, both are called on the game thread so it's safe to edit attributes here without synchronization.

plugin-dev/Source/Sentry/Private/Utils/SentryPerformanceConsumer.cpp
tustanivsky: Hardware info (`gpu.name`, `cpu.cores`, `ram.gb`) is truly static. Resolution could theoretically change but I think we can treat it as static as well since most games run fullscreen anyway. The `map` attribute is the only dynamic one but it's handled separately via the `PostLoadMapWithWorld` delegate.

jpnurmi (approved):

PR #1325, opened 2026-03-31
Add out-of-process screenshot capturing on Windows

tustanivsky:
The current in-process Slate-based screenshot capture during crashes might now always work because the rendering pipeline may be in a broken state when the crash occurs. The native SDK's out-of-process approach captures screenshots from a separate process making it significantly more reliable. It also supports fast-fail crashes that bypass SEH (crashpad backend).

### Summary

- Adds a new `EnableOutOfProcessScreenshots` setting under `General|Attachments` that delegates crash screenshot capture to sentry-native's `sentry_options_set_attach_screenshot` API
- When enabled on Windows, crash screenshots are captured out-of-process (via crashpad handler or native daemon) using GDI+ instead of in-process via Unreal's Slate renderer
- Non-crash screenshots (ensures, hangs) continue using the existing Slate-based capture on all platforms
- Default is off (existing Slate-based behavior is preserved unless explicitly opted in)

### Documentation
- https://github.com/getsentry/sentry-docs/pull/17208

### DIFF
--- a/plugin-dev/Source/Sentry/Public/SentrySettings.h
+++ b/plugin-dev/Source/Sentry/Public/SentrySettings.h
@@ -319,10 +319,10 @@ class SENTRY_API USentrySettings : public UObject
 	UPROPERTY(Config, EditAnywhere, Category = "General|Attachments",
 		Meta = (DisplayName = "Attach screenshots", ToolTip = "Flag indicating whether to attach screenshot of the application when an error occurs. Currently this feature is supported for Windows and Linux only."))
 	bool AttachScreenshot;
-	
+
 	UPROPERTY(Config, EditAnywhere, Category = "General|Attachments",
-	Meta = (DisplayName = "Enable out-of-process screenshot capturing (for Windows only, experimental)", ToolTip = "Flag indicating whether crash screenshots should be captured out-of-process by the native SDK using OS-level APIs instead of Unreal's Slate renderer. More reliable during crashes since it doesn't depend on the rendering pipeline being functional. Currently supported on Windows only. Requires `Attach screenshots` to be enabled.",
-		EditCondition = "AttachScreenshot"))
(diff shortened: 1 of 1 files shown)


plugin-dev/Source/Sentry/Public/SentrySettings.h
```diff
@@ -467,6 +467,11 @@ class SENTRY_API USentrySettings : public UObject
 		Meta = (DisplayName = "Enable logging during crash handling", ToolTip = "Flag indicating whether the SDK should log additional crash information (such as stack traces and error messages). This is intended for debug builds only and is not safe for production use."))
 	bool EnableOnCrashLogging;
 
+	UPROPERTY(Config, EditAnywhere, Category = "General|Native",
+		Meta = (DisplayName = "Out-of-process screenshot capturing (Experimental)", ToolTip = "Flag indicating whether crash screenshots should be captured out-of-process by the native SDK using OS-level APIs instead of Unreal's Slate renderer. More reliable during crashes since it doesn't depend on the rendering pipeline being functional. Currently supported on Windows only. Requires `Attach screenshots` to be enabled.",
+			EditCondition = "AttachScreenshot"))
+	bool UseNativeScreenshotCapturing;
+
```
bitsandfoxes: I think we should either bundle "screenshot capture" into one option/flag, or if this is split into two because it is experimental, have `experimental` in the name. Either that or there should be a hierarchy to them. The user opts into screenshot capture, and then also opts into capturing them out-of-process.

Documenting and explaining this here and on other platforms is going to be really hard because when is what captured by who? In which error case is what option going to kick in?

tustanivsky: > Either that or there should be a hierarchy to them. The user opts into screenshot capture, and then also opts into capturing them out-of-process

That make sense - I'll put this new setting next to `AttachScreenshot` so it will be easier for users to configure all screenshot-related matters in one place. We can also benefit from covering this in more details docs.

tustanivsky: Follow-up docs PR: https://github.com/getsentry/sentry-unreal/pull/1325

tustanivsky: Also, grouped two settings together:
image

PR #1341, opened 2026-04-14
Add consent-aware offline caching

### LINKED ISSUE #1103: Allow sentry to init in "offline mode" so it doesn't communicate with the internet
We have software which can be booted in "offline mode". The expectation is that, even if there is an internet connection, we do not communicate with the internet. When sentry is initialized, it makes an HTTP call, regardless if it will be used to report crashes. 

What we desire is a way to init sentry without making any http calls. Any crashes collected by sentry are left on disk while we are "offline", and when the software is restarted in "online mode" any crashes which are found on disk are then uploaded to the sentry portal. 

To my knowledge, if we use a to disk transport, we can get the crashes being written to disk without uploading them to the sentry portal, but I don't know of a way to init sentry without it attempting to make an HTTP call.

tustanivsky:
This PR enables HTTP retries unconditionally during SDK initialization on all sentry-native platforms where HTTP transport is used

It is required for consent-aware offline caching (https://github.com/getsentry/sentry-native/pull/1542) - when consent is revoked and offline caching is enabled, envelopes are cached to disk. Once consent is granted again, triggers the retry system to immediately flush cached envelopes.

Closes #1103

Depends on:
- https://github.com/getsentry/sentry-native/pull/1542


plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentrySubsystem.cpp
```diff
@@ -530,6 +530,7 @@ void FGenericPlatformSentrySubsystem::InitWithSettings(const USentrySettings* se
 	sentry_options_set_logs_with_attributes(options, true);
 	sentry_options_set_enable_metrics(options, settings->EnableMetrics);
 	sentry_options_set_before_send_metric(options, HandleBeforeMetric, this);
+	sentry_options_set_http_retry(options, 1);
```
jpnurmi, line 533: Makes perfect sense because this brings the Native platforms on par with the Android and Cocoa siblings that have unconfigurable retries. :+1:

jpnurmi (approved):

Issue #1396, opened 2026-05-21
iOS Simulator build fails: Sentry.embeddedframework.zip is device-only


microneo:
### Environment

How do you use Sentry?
Self-hosted

Which version of the SDK?
1.12.0

How did you install the package?
Git-URL

Which version of Unreal?
5.6/5.7

Is this happening in Unreal (editor) or on a player like Android, iOS, Windows?
iOS Simulator

### Steps to Reproduce

1. Add Sentry plugin to an Unreal 5.7 project
2. Build for iOS Simulator:
   `Engine/Build/BatchFiles/Mac/Build.sh ProjectName IOS Development -Project=ProjectName.uproject -Architecture=iossimulator`

### Expected Result

Project compiles successfully for iOS Simulator.

### Actual Result

Linker error:
`ld: building for 'iOS-simulator', but linking in dylib
(.../UnzippedFrameworks/Sentry/Sentry.embeddedframework/Sentry.framework/Sentry)
built for 'iOS'
clang++: error: linker command failed with exit code 1`

`Sentry.embeddedframework.zip` contains only an `arm64 iphoneos` slice - no simulator slice is included.

A fix is already prepared: https://github.com/microneo/sentry-unreal/commit/4ef048d7ffc829a2247cf09a7359b8c0ecd0130f

Happy to open a PR if this approach looks good.

tustanivsky:
@microneo Thank you for raising this - we'll review the suggested changes and get back to you.

microneo:
I made some minor changes, and it builds successfully on UE 5.7.

However, there is still one more issue: UE5 has two build bugs for the iOS simulator. I’m working on fixing them as well. They are not major, and I’ll try to provide a patch for UE 5.7 today that fixes this problem. I will also submit a PR to Epic Games.

microneo:
[ios_simulator_fix_5.7.patch](https://github.com/user-attachments/files/28163422/ios_simulator_fix_5.7.patch)

PR #1403, opened 2026-05-22
Fix iOS Simulator build for UE 5.6+

microneo:
**Changelog:**

* Switch to Sentry.xcframework.zip containing both device and simulator slices
* Keep Sentry.embeddedframework.zip with UPL embedding for older UE versions

**Description:**

When building an Unreal project for iOS Simulator, the build fails because
`Sentry.embeddedframework.zip` contains only an `arm64 iphoneos` slice with no
simulator slice included.

**Build command:**

`Engine/Build/BatchFiles/Mac/Build.sh ProjectName IOS Development -Project=ProjectName.uproject -Architecture=iossimulator`

**Error:**

`ld: building for 'iOS-simulator', but linking in dylib
  (.../UnzippedFrameworks/Sentry/Sentry.embeddedframework/Sentry.framework/Sentry)
  built for 'iOS'
clang++: error: linker command failed with exit code 1`

#1396

### DIFF
--- a/plugin-dev/Source/Sentry/Sentry.Build.cs
+++ b/plugin-dev/Source/Sentry/Sentry.Build.cs
@@ -62,11 +62,14 @@ public class Sentry : ModuleRules
 		{
 			PrivateIncludePaths.Add(Path.Combine(ModuleDirectory, "Private", "Apple"));
 
+#if UE_5_6_OR_LATER
+			PublicAdditionalFrameworks.Add(new Framework("Sentry", Path.Combine(PlatformThirdPartyPath, "Sentry.xcframework.zip"), null, true));
+#else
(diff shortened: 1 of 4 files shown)


microneo:
Unfortunately, after integrating `Sentry.xcframework.zip` from this PR, I ran into an TestFlight rejection:

```text
ITMS-90338: Non-public API usage.
The app references non-public selectors in <AppName>: removeDataForKey:
```

I opened a corresponding issue in `sentry-cocoa`:

https://github.com/getsentry/sentry-cocoa/issues/7942

It looks like `removeDataForKey:` comes from Sentry Cocoa 9.13.0 and is present in the final binary when the new `Sentry.xcframework` is bundled with the Unreal plugin.

The simulator fix works, but this may block TestFlight/App Store uploads until `sentry-cocoa` renames that Objective-C selector or provides a fixed XCFramework.

tustanivsky:
Superseded by #1561

PR #1420, opened 2026-06-03
Fix Sentry Crash Reporter showing an empty window for packaged games on Mac

tustanivsky:
This PR fixes the Sentry Crash Reporter showing an empty window (no crash details) when the game runs under the macOS App Sandbox (packaged game build).

`MacSentrySubsystem::GetCrashReporterExecutableName()` now returns the path to the executable inside the bundle (`Sentry.CrashReporter.app/Contents/MacOS/Sentry.CrashReporter`) instead of the `.app` directory. sentry-native's process spawner branches on the `.app` suffix - with the direct binary path the daemon spawns the Crash Reporter via `execv` rather than `/usr/bin/open`. The Crash Reporter then inherits the game's sandbox at the kernel level and can read the crash envelope inside the parent's container natively with no TCC permission prompt (which wouldn't appear during a crash anyway) and no cross-container access denial.

Also updates the sample's `DefaultEngine.ini` to override `ShippingSpecificMacEntitlements` so the playground's Shipping packaged builds retain `com.apple.security.network.client` (Unreal's default Shipping template is  `Sandbox.NoNet.entitlements`).

Related items:
- https://github.com/getsentry/sentry-docs/pull/18010

### DIFF
--- a/plugin-dev/Source/Sentry/Private/Mac/MacSentrySubsystem.h
+++ b/plugin-dev/Source/Sentry/Private/Mac/MacSentrySubsystem.h
@@ -19,7 +19,7 @@ protected:
 	virtual void ConfigureCrashReporterPath(sentry_options_t* Options) override;
 
 	virtual FString GetHandlerExecutableName() const override;
-	virtual FString GetCrashReporterExecutableName() const override { return TEXT("Sentry.CrashReporter.app"); }
+	virtual FString GetCrashReporterExecutableName() const override { return TEXT("Sentry.CrashReporter.app/Contents/MacOS/Sentry.CrashReporter"); }
 
 	virtual bool IsScreenshotSupported() const override { return true; }
 	virtual bool IsHangTrackingSupported() const override { return false; }
--- a/sample/Config/DefaultEngine.ini
+++ b/sample/Config/DefaultEngine.ini
@@ -242,4 +242,5 @@ UploadSymbolsAutomatically=True
 [/Script/MacTargetPlatform.XcodeProjectSettings]
 BundleIdentifier=io.sentry.unreal.sample
 CodeSigningPrefix=
+ShippingSpecificMacEntitlements=(FilePath="/Game/Build/Mac/Resources/Sandbox.Server.entitlements")
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@
 - Add ANR timeout setting for mobile platforms ([#1413](https://github.com/getsentry/sentry-unreal/pull/1413))
 - Enable ANR error capturing by default on Android and Apple platforms ([#1415](https://github.com/getsentry/sentry-unreal/pull/1415))
 
+### Fixes
+
+- Fix Sentry Crash Reporter showing an empty window for packaged games on Mac ([#1420](https://github.com/getsentry/sentry-unreal/pull/1420))
+
 ### Dependencies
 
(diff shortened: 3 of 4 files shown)


sample/Config/DefaultEngine.ini
```diff
@@ -242,4 +242,5 @@ UploadSymbolsAutomatically=True
 [/Script/MacTargetPlatform.XcodeProjectSettings]
 BundleIdentifier=io.sentry.unreal.sample
 CodeSigningPrefix=
+ShippingSpecificMacEntitlements=(FilePath="/Game/Build/Mac/Resources/Sandbox.Server.entitlements")
```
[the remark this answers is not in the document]
tustanivsky: The file is intentionally not vendored as UnrealBuildTool auto-copies it from the engine's resource folder at packaging time and resolves the correct path automatically.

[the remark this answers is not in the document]
tustanivsky: The file is intentionally not vendored as UnrealBuildTool auto-copies it from the engine's resource folder at packaging time and resolves the correct path automatically.

plugin-dev/Source/Sentry/Private/Mac/MacSentrySubsystem.h
```diff
@@ -19,7 +19,7 @@ class FMacSentrySubsystem : public FGenericPlatformSentrySubsystem
 	virtual void ConfigureCrashReporterPath(sentry_options_t* Options) override;
 
 	virtual FString GetHandlerExecutableName() const override;
-	virtual FString GetCrashReporterExecutableName() const override { return TEXT("Sentry.CrashReporter.app"); }
+	virtual FString GetCrashReporterExecutableName() const override { return TEXT("Sentry.CrashReporter.app/Contents/MacOS/Sentry.CrashReporter"); }
```
jpnurmi, line 22: Will this also change it for non-sandboxed games? Would it be possible to detect a sandboxed environment and only resort to the direct executable path if so?

The main reason we used `open -a` for bundles was to ensure that the crash dialog is appropriately activated and raised on top. With `execv`, the crash dialog would open stacked below the crashed app window. 😭

jpnurmi: Actually, we could even consider having such detection logic in the Native SDK, so it works out of the box for all downstream SDKs and other Native SDK users... What do you think? 🤔

tustanivsky: > Will this also change it for non-sandboxed games?

Yes, in case with Unreal that would be the `Editor` target. Packaged game builds are sandboxed by default.

> Would it be possible to detect a sandboxed environment and only resort to the direct executable path if so?

We can use Apple's Security framework to check whether the app has `com.apple.security.app-sandbox` entitlement via [SecTaskCopyValueForEntitlement](https://developer.apple.com/documentation/security/sectaskcopyvalueforentitlement(_:_:_:)) which allows to determine the environment.

However, I haven't noticed anything unusual about how the Crash Reporter dialog is displayed so far when the game crashes - the game window disappears and the Crash Reporter window pops up as expected.

So given that, I'm not sure there's a need for separate code paths for sandboxed and non-sandboxed apps 🤔

mujacica (approved):

jpnurmi (approved):

PR #1496, opened 2026-07-24
chore: Harden integration-test crash uploads in CI

tustanivsky:
The integration tests were flaky in CI for several distinct reasons - each confirmed from the failure artifacts this PR started capturing:

- **Shutdown upload races** - non-crash tests (message / log / metric / tracing) upload through the SDK transport on `sentry_close`, bounded by `ShutdownTimeout` (default **5000 ms**). Large envelopes or a slow link can exceed that window and the in-flight upload gets cancelled. The same drain-on-shutdown bound matters for the sentry-xbox extension, so we want generous headroom here regardless of the desktop legs. On the crashpad legs, the crashing process can also exit before the out-of-process handler finishes its upload unless it is configured to wait.
- **OutOfMemory, Windows native backend** - the OOM test exhausted the whole runner's memory, so the `sentry-crash` daemon captured the crash but sometimes failed to upload the envelope (`WinHttpSendRequest` error 8, `ERROR_NOT_ENOUGH_MEMORY`); with no retry or durable fallback the event was lost.
- **StackOverflow, Windows, both backends** - no thread stack guarantee was in effect due to sentry-native bugs, so the in-process crash handler had no stack to run on and no minidump was created.
- **Android crash test (SauceLabs)** - real devices emit thousands of unrelated logcat lines per second (in one failing capture, 7.8k of the 10k captured lines were `AuthPII` spam spanning 1.2 s), evicting UE's `EVENT_CAPTURED` marker before the harness reads the log. The crash itself succeeded.
- **Replay duration assertion** - the server derives replay duration from segment timestamps, which shift with crash-handling timing, so an exact-match assertion is inherently flaky.

## Key Changes

**`sample/Config/DefaultEngine.ini`**

- `ShutdownTimeout=30000` - gives the native daemon's crash upload and all async (non-crash) event uploads time to finish on shutdown before the transport is torn down. Drains early when the queue empties, so there is no slowdown when uploads are fast.
- `CrashpadWaitForUpload=True` - on the crashpad legs, the process waits for the crashpad upload to complete before exiting, so the crash is reliably delivered during the crash run.

**`integration-test-windows.yml` / `integration-test-linux.yml`**

- Include logs from `sample-build/SentryPlayground/.sentry-native/` in the on-failure artifacts so the crash DB (minidumps, pending/completed uploads, daemon logs) is available when a test fails.

**`integration-test-android.yml`**

- `SAUCE_LOGCAT_FILTER` (via https://github.com/getsentry/app-runner/pull/53, already in the pinned revision) captures only `UE`/`Sentry`/tombstone tags so app markers can't be evicted by device spam.

**`SentryPlaygroundTestUtils.{h,cpp}, SentryCrashTest.cpp`**

- New `SetMemoryLimit(HeadroomMB)` util: the process self-assigns a Job Object with `JOB_OBJECT_LIMIT_PROCESS_MEMORY` = current commit + headroom (Windows-only, no-op elsewhere). The OOM test calls it with 4 GB, so allocation fails in-process while the runner (and the crash daemon) keep full memory headroom. Also drops the OOM test from ~6.5 min of pagefile thrashing to ~7 s.

**`Integration.Desktop.Tests.ps1`**

- Dropped the exact replay-duration assertion; the server derives duration from segment timestamps, which shift with crash-handling timing.

## Notes

- `ShutdownTimeout` and the `.sentry-native` artifact help both the crashpad and native desktop legs; `CrashpadWaitForUpload` is a no-op on the native legs, so it is safe there.
- The `.sentry-native` artifact path assumes `DatabaseLocation=ProjectUserDirectory` resolves to the packaged project dir (sibling of `Saved/Logs/`).

## Related Items

- https://github.com/getsentry/sentry-native/pull/1918
- https://github.com/getsentry/sentry-unreal/pull/1500

#skip-changelog

### DIFF
--- a/.github/workflows/integration-test-windows.yml
+++ b/.github/workflows/integration-test-windows.yml
@@ -52,3 +52,5 @@ jobs:
             integration-test/output/
             sample-build/SentryPlayground/Saved/Logs/
             sample-build/SentryPlayground/.sentry-native/
+          # .sentry-native is a hidden dir; upload-artifact skips hidden files by default
+          include-hidden-files: true
--- a/.github/workflows/integration-test-windows.yml
+++ b/.github/workflows/integration-test-windows.yml
@@ -51,3 +51,4 @@ jobs:
           path: |
             integration-test/output/
             sample-build/SentryPlayground/Saved/Logs/
+            sample-build/SentryPlayground/.sentry-native/
--- a/.github/workflows/integration-test-windows.yml
+++ b/.github/workflows/integration-test-windows.yml
@@ -51,6 +51,5 @@ jobs:
           path: |
             integration-test/output/
             sample-build/SentryPlayground/Saved/Logs/
-            sample-build/SentryPlayground/.sentry-native/
-          # .sentry-native is a hidden dir; upload-artifact skips hidden files by default
+            sample-build/SentryPlayground/.sentry-native/**/*.log
           include-hidden-files: true
--- a/.github/workflows/integration-test-windows.yml
+++ b/.github/workflows/integration-test-windows.yml
@@ -39,8 +39,6 @@ jobs:
           SENTRY_UNREAL_TEST_DSN: ${{ secrets.SENTRY_UNREAL_TEST_DSN }}
           SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_API_TOKEN }}
           SENTRY_UNREAL_TEST_APP_PATH: ${{ github.workspace }}/sample-build/SentryPlayground.exe
-          # Enlarged thread stack guarantee (KiB) so the crash handler can run on stack-overflow (default 64 is not always enough for UE's handler chain)
-          SENTRY_HANDLER_STACK_SIZE: 256
         run: |
           cmake -B build -S .
           Invoke-Pester Integration.Desktop.Tests.ps1 -CI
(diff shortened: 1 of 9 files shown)


.github/workflows/integration-test-windows.yml
```diff
@@ -51,3 +51,5 @@ jobs:
           path: |
             integration-test/output/
             sample-build/SentryPlayground/Saved/Logs/
+            sample-build/SentryPlayground/.sentry-native/**/*.log
```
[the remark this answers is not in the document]
tustanivsky: That's expected at the moment (see https://github.com/getsentry/sentry-unreal/pull/1496#discussion_r3663855539).

integration-test/Integration.Desktop.Tests.ps1
```diff
@@ -615,11 +615,6 @@ Describe "Sentry Unreal Desktop Integration Tests (<Platform>)" -ForEach $TestTa
             $script:ReplayEvent.count_segments | Should -Be 1
         }
 
-        It "Should have expected replay duration" {
-            # The staged clip's sidecar claims a 5000 ms duration
-            $script:ReplayEvent.duration | Should -Be 5
-        }
-
```
JoshuaMoelans, line 622: q: does it make sense to fully remove the assertion, or could we simply weaken it (e.g., > 3)? maybe not necessary since we already check for the segment count above though

JoshuaMoelans (approved):
LGTM, just small q on the Replay Duration

integration-test/Integration.Desktop.Tests.ps1
tustanivsky: Given that we can't reliably determine replay duration here I'd say we drop this assertion entirely. Probably segment count check is sufficient for now.

PR #1497, opened 2026-07-24
Add app-hang tracking via NDK game-thread heartbeat on Android

tustanivsky:
This PR adds app-hang detection on Android that monitors the Unreal game thread. Until now, Android only had the JVM ANR watchdog (`enableAnrTracking`), which watches the Android main/UI looper - a different thread from Unreal's game thread. A frozen game loop while the UI thread keeps pumping went undetected. This wires up sentry-native's in-process app-hang watchdog on Android (exposed to hybrid SDKs via [getsentry/sentry-java#5623](https://github.com/getsentry/sentry-java/pull/5623)), giving the same game-thread hang detection that is already available on Windows/macOS/Linux - with a native stack trace and consistent event shape across platforms.

The JVM ANR watchdog and the new NDK app-hang watchdog both stay active: they watch different threads and catch different failure modes (system ANR vs. frozen game loop).

## Key Changes

- `AndroidSentrySubsystem` - forwards `enableNdkAppHangTracking` / `ndkAppHangTimeoutMillis` to the Java layer, gated on `EnableHangTracking && UseNativeHangTracking`. Pumps `sentry_app_hang_heartbeat()` from `FCoreDelegates::OnEndFrame` (game thread latches itself as the monitored thread), resolving the symbol lazily via `dlopen("libsentry.so", RTLD_NOLOAD)` + `dlsym` against the library already loaded by the Java SDK. `IsNativeHangTrackingEnabled()` now reflects the feature; `IsHangTrackingSupported()` stays false so the engine `FThreadHeartBeat` watcher is never created on Android.
- `SentryBridgeJava` - applies `setEnableNdkAppHangTracking` / `setNdkAppHangTimeoutIntervalMillis` on `SentryAndroidOptions`.
- `SentrySettings` - reuses the existing hang-tracking settings (no new options); updated the `UseNativeHangTracking` tooltip to note Android coverage.

## Documentation

- https://github.com/getsentry/sentry-docs/pull/18828

## Related items

- https://github.com/getsentry/sentry-java/pull/5623
- https://github.com/getsentry/sentry-unity/pull/2745
- https://github.com/getsentry/sentry-unreal/pull/1427

### DIFF
--- a/plugin-dev/Source/Sentry/Private/Android/AndroidSentrySubsystem.cpp
+++ b/plugin-dev/Source/Sentry/Private/Android/AndroidSentrySubsystem.cpp
@@ -35,6 +35,8 @@
 #include "Serialization/JsonSerializer.h"
 #include "Utils/SentryScreenshotUtils.h"
 
+#include <dlfcn.h>
+
 FAndroidSentrySubsystem::FAndroidSentrySubsystem()
 {
 	SentryJavaClasses::InitJavaClassRefsCache();
@@ -55,6 +57,8 @@ void FAndroidSentrySubsystem::InitWithSettings(const USentrySettings* settings, 
 
 	isScreenshotAttachmentEnabled = settings->AttachScreenshot;
 
+	bNdkAppHangTracking = settings->EnableHangTracking && settings->UseNativeHangTracking;
+
 	TSharedPtr<FJsonObject> SettingsJson = MakeShareable(new FJsonObject);
 	SettingsJson->SetStringField(TEXT("dsn"), settings->Dsn);
 	SettingsJson->SetStringField(TEXT("release"), settings->GetEffectiveRelease());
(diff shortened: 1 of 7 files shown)


plugin-dev/Source/Sentry/Private/Android/AndroidSentrySubsystem.cpp
```diff
@@ -73,6 +77,8 @@ void FAndroidSentrySubsystem::InitWithSettings(const USentrySettings* settings,
 	SettingsJson->SetBoolField(TEXT("sendDefaultPii"), settings->SendDefaultPii);
 	SettingsJson->SetBoolField(TEXT("enableAnrTracking"), settings->EnableAppNotRespondingTracking);
 	SettingsJson->SetNumberField(TEXT("anrTimeoutMillis"), settings->AppNotRespondingTimeout * 1000.0f);
+	SettingsJson->SetBoolField(TEXT("enableNdkAppHangTracking"), bNdkAppHangTracking);
+	SettingsJson->SetNumberField(TEXT("ndkAppHangTimeoutMillis"), settings->HangTimeoutDuration * 1000.0f);
```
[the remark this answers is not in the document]
tustanivsky: `enableNdk` setting is unrelated to app-hang tracking.

plugin-dev/Source/Sentry/Private/Android/AndroidSentrySubsystem.cpp
```diff
@@ -525,3 +548,27 @@ FString FAndroidSentrySubsystem::TryCaptureScreenshot() const
 
 	return ScreenshotPath;
 }
+
+void FAndroidSentrySubsystem::PumpAppHangHeartbeat()
+{
+	if (!AppHangHeartbeatFunc)
+	{
+		// libsentry.so is loaded asynchronously by the Java SDK (io.sentry.ndk.SentryNdk), so resolve
+		// the symbol lazily against the already-loaded library rather than linking it at build time.
```
jpnurmi, line 557: does this mean that it can take several frames on startup until it succeeds to resolve the symbol? does it have any measurable impact on the startup frames?

tustanivsky: In practice it resolves on the very first frame, so no. The `OnEndFrame `listener is only registered after the Android SDK initialization completes and that initialization waits for the NDK library to finish loading.

jpnurmi: Sorry for the silly questions, but do we have any mechanism for delaying app hang tracking startup? Some games might want to start it only after level loading and other startup work is complete...

jpnurmi (approved):

mujacica (approved):

plugin-dev/Source/Sentry/Private/Android/AndroidSentrySubsystem.cpp
tustanivsky: App hang tracking starts with the first heartbeat so technically we could expose a setting to control whether it happens automatically after init along with an API to trigger it manually later.

It'd probably also be useful to support pausing and resuming hang tracking since level loading (or other long-running operations) can happen multiple times during a game session.

cc @bitsandfoxes

PR #1510, opened 2026-07-31
Defer session replay until engine initialization

ac-branislav:
## Summary

- Defer session replay startup until the engine loop has completed initialization.
- Remove the deferred startup delegate after it fires and during subsystem shutdown.
- Verify that the hardware device has an RHI video context before creating the encoder.

## Testing

- Built SentryPlaygroundEditor with session replay enabled using Unreal Engine 5.7.4.
- Passed all 66 Sentry automation tests.
- Completed a D3D12 and NVENC replay run and verified the encoder opened after engine initialization.
- Verified valid fragmented MP4 output and a clean shutdown.

### DIFF
--- a/plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentrySubsystem.cpp
+++ b/plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentrySubsystem.cpp
@@ -39,6 +39,7 @@
 #include "GenericPlatform/CrashReporter/GenericPlatformSentryCrashContext.h"
 #include "GenericPlatform/CrashReporter/GenericPlatformSentryCrashReporter.h"
 
+#include "CoreGlobals.h"
 #include "Dom/JsonObject.h"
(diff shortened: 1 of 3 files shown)


grujicbr:
I realized I forgot to add a comment for this but I could never launch our game without this, I'm a bit confused *how* this ever worked, maybe because we are using DX12? But UE5 uses DX12 by default so still slightly confused

tustanivsky:
Although it's a bit surprising that the issue above never came up during dev testing the suggested changes make sense. Based on the context in #1509, this looks like a race between engine initialization and the plugin's replay capturer, so deferring the latter until the post-init stage is a good call regardless.

tustanivsky (approved):

tustanivsky:
Applied similar patch for macOS/iOS in https://github.com/getsentry/sentry-unreal/pull/1514

ac-branislav:
I think it might depend if you boot right into a world or if you have a loading screen. In our case we have a loading screen that shows up before the main menu comes up, but yah either way it's a good change

PR #1511, opened 2026-07-31
Resolve writable Sentry database path

ac-branislav:
## Summary

- Resolve the Sentry database directory with ConvertToAbsolutePathForExternalAppForWrite.
- Keep replay output in the same writable location used by the native SDK attachment path.

## Testing

- Built SentryPlaygroundEditor with session replay enabled using Unreal Engine 5.7.4.
- Passed all 66 Sentry automation tests.
- Completed a D3D12 and NVENC replay run using the resolved absolute path.
- Verified a valid 150-frame fragmented MP4 was written successfully.

### DIFF
--- a/plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentrySubsystem.cpp
+++ b/plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentrySubsystem.cpp
@@ -1348,7 +1348,7 @@ FString FGenericPlatformSentrySubsystem::GetCrashReporterLogoPath() const
 FString FGenericPlatformSentrySubsystem::GetDatabasePath() const
 {
 	const FString DatabasePath = FPaths::Combine(databaseParentPath, TEXT(".sentry-native"));
-	const FString DatabaseFullPath = FPaths::ConvertRelativePathToFull(DatabasePath);
+	const FString DatabaseFullPath = IFileManager::Get().ConvertToAbsolutePathForExternalAppForWrite(*DatabasePath);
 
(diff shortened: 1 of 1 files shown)


tustanivsky:
@ac-branislav Did you observe any actual write failures which required this fix? On desktop, `ConvertToAbsolutePathForExternalAppForWrite` just forwards to `FPaths::ConvertRelativePathToFull` which is the same thing the original code does and thus it resolves to an identical path:

https://github.com/EpicGames/UnrealEngine/blob/260bb2e1c5610b31c63a36206eedd289409c5f11/Engine/Source/Runtime/Core/Private/GenericPlatform/GenericPlatformFile.cpp#L998-L1001

It'd only differ on consoles that override this method which isn't the case for the platforms this path serves.

ac-branislav:
I ran into this on PC, i simple cooked a package using compressed .pak files and than when I ran that build the log was reporting the path is where sentry reads it from versus where it writes it to, so there is a missmatch there for sure. Once i added this, all of the cases worked

tustanivsky:
> cooked a package using compressed .pak files

Thanks, that might actually be the culprit - we'll try to repro on our end with a compressed-pak build and let you know.

tustanivsky:
We confirmed that when Unreal remaps write paths (e.g. sandbox mode) the plugin ignores it and this potentially can cause issues with sentry-native's database creation.

Since `ConvertToAbsolutePathForExternalAppForWrite` is strictly safer (identical path when no remap is active, correct writable path when one is) we're good to merge this.

tustanivsky (approved):

PR #1523, opened 2026-08-06
Pause native app-hang detection while the app is backgrounded

tustanivsky:
App-hang detection feeds a heartbeat from the game thread's `OnEndFrame` and reports a hang if the watchdog doesn't see one within the timeout. When the OS backgrounds or suspends the app, the monitored thread stops ticking through no fault of the game, so the watchdog fires a **false-positive** app-hang event. This PR pauses hang detection across those app-state transitions and resumes it when the app returns.

It builds on the approach proposed in #1506 (feeding the native heartbeat from `OnEndFrame` and pausing via `sentry_app_hang_pause`), extended so the pause/resume wiring can be platform-specific.

### Platform-specific behavior

The app-lifecycle delegates that signal background/suspend are broadcast very differently depending on platform (full analysis in https://github.com/getsentry/sentry-unreal/pull/1506#issuecomment-5202892027):

- **Android** - the deactivate/reactivate and background/foreground pairs are always broadcast together, so a single pair is enough.
- **Desktop** (Win/Linux/Mac) - the engine never broadcasts these delegates, so no pausing is needed and none should be wired.
- **PS5** - only the deactivate/reactivate pair is used (`WillEnterBackground` is never broadcast).
- **Xbox** - both pairs fire, for different scenarios: deactivate/reactivate = *constrained* (still ticking but throttled - the main false-positive case), background/foreground = *suspend/resume*.
- **Switch** - both pairs are used, but the engine restores only the delegate matching the most recent focus state on resume, so a naive two-flag approach can get stuck paused after a mixed transition (e.g. overlay → HOME → resume leaves the earlier deactivation unmatched).

No single shared implementation satisfies all of these, so the generic-platform layer exposes overridable hooks and each platform supplies its own pause policy.

### Key Changes

- **GenericPlatform:** app-hang wiring is exposed as three overridable hooks - `ConfigureAppHangTracking()` (subscribes the `OnEndFrame` heartbeat), `ResetAppHangTracking()` (removes it), and `IsAppHangTrackingActive()` (gates the heartbeat; `true` by default). The base only pumps the heartbeat.
- **Android:** subscribes only the background/foreground pair and tracks a single foreground flag.
- **Windows:** overrides the hooks back to the heartbeat-only default, since the lifecycle delegates never fire on desktop.
- **Xbox (XB1/XSX/WinGDK):** implemented in the shared Microsoft subsystem - both pairs, gated on being active **and** foregrounded.

### Downstream extensions

The console extensions live in separate repos and each provides its own pause policy, so they need matching patches:
- PS5 → getsentry/sentry-playstation#156
- Switch → getsentry/sentry-switch#169

Xbox needs no separate extension patch - its targets inherit the shared Microsoft subsystem changed here.


plugin-dev/Source/Sentry/Private/Microsoft/MicrosoftSentrySubsystem.cpp
```diff
@@ -146,6 +147,64 @@ void FMicrosoftSentrySubsystem::Close()
...
+		sentry_app_hang_pause();
+	});
+	AppHangHasEnteredForegroundHandle = FCoreDelegates::ApplicationHasEnteredForegroundDelegate.AddLambda([this]()
+	{
+		bAppIsForeground.AtomicSet(true);
+	});
+	AppHangHasReactivatedHandle = FCoreDelegates::ApplicationHasReactivatedDelegate.AddLambda([this]()
+	{
+		bAppIsActive.AtomicSet(true);
+	});
```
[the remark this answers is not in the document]
tustanivsky: sentry-native has no `sentry_app_hang_resume()`; a paused detector resumes automatically on the next `sentry_app_hang_heartbeat()` from the watched thread. On foreground/reactivate we set the flags true, so the next `OnEndFrame` pumps a heartbeat and detection is live again.

bitsandfoxes (approved):
Thanks for carrying this over the finish line! LGTM!

plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentrySubsystem.cpp
JoshuaMoelans: we still only have bNativeHangTracking for WINDOWS/MAC/LINUX being set here, would we need to extend these for PlayStation/XBOX/Switch too?

tustanivsky: That's a good catch - for Xbox support we can actually loosen `PLATFORM_WINDOWS` to `PLATFORM_MICROSOFT`. However, for Switch and PS5 let’s leave it as-is for now since despite having all the required plumbing on the Unreal side for these platforms native hang tracking is currently a no-op there.

JoshuaMoelans: alright sounds good 👌

JoshuaMoelans (approved):

PR #1542, opened 2026-08-19
Stop overwriting native-collected device context on consoles

tustanivsky:
This PR fixes device context data loss on platforms where the native SDK (or its console extensions) pre-populates the `device` context on the scope at init.

`AddDeviceContext` previously used `SetContext`, which fully replaces the context object under a key. On the game consoles (PS5, Xbox, Switch), whose extension layers seed `device` with `model`/`family`/`manufacturer`/`arch` during init, this overwrote the SDK-collected fields with the plugin's generic subset, losing the console device identity.

A new internal `UpdateContext` API merges the provided values into the existing context (new values take precedence) instead of replacing it, so both the SDK-collected fields and the Unreal-specific ones survive.

## Key Changes

- Add `UpdateContext` to the subsystem interface with per-platform implementations:
  - **Native / consoles**: `sentry_update_context` (the sentry-native merge primitive).
  - **Cocoa**: emulated read-merge-write on the scope (sentry-cocoa has no merge primitive).
  - **Android**: scope-level read-merge-write via a new `SentryBridgeJava.updateContext` (benefits custom user contexts).
  - **Null**: no-op stub.
- Route `AddDeviceContext` through `UpdateContext` so console/desktop device identity is preserved alongside the Unreal-specific fields.

Note: Setting the `device` context is effectively a no-op on Android: sentry-java rebuilds the `device` context per event with richer device info (manufacturer, model, family, arch, chipset, processor count, memory, screen, …) and overwrites the scope value at capture time, while the Unreal-specific `device_type` is merged back in via `beforeSend`. The new `UpdateContext` still applies for custom user contexts on Android.

## Related Items

- getsentry/sentry-native#1835

### DIFF
--- a/plugin-dev/Source/Sentry/Private/Apple/AppleSentrySubsystem.cpp
+++ b/plugin-dev/Source/Sentry/Private/Apple/AppleSentrySubsystem.cpp
@@ -592,6 +592,19 @@ void FAppleSentrySubsystem::SetContext(const FString& key, const TMap<FString, F
 	}];
 }
 
+void FAppleSentrySubsystem::UpdateContext(const FString& key, const TMap<FString, FSentryVariant>& values)
+{
+	[SENTRY_APPLE_CLASS(SentryObjCSDK) configureScope:^(SentryObjCScope* scope) {
+		NSDictionary* contexts = [scope serialize][@"context"];
+		NSDictionary* existing = [contexts objectForKey:key.GetNSString()];
+
+		NSMutableDictionary* merged = existing ? [existing mutableCopy] : [NSMutableDictionary dictionary];
+		[merged addEntriesFromDictionary:FAppleSentryConverters::VariantMapToNative(values)];
+
+		[scope setContextValue:merged forKey:key.GetNSString()];
+	}];
+}
+
 void FAppleSentrySubsystem::SetTag(const FString& key, const FString& value)
 {
 	[SENTRY_APPLE_CLASS(SentryObjCSDK) configureScope:^(SentryObjCScope* scope) {
--- a/plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentrySubsystem.cpp
+++ b/plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentrySubsystem.cpp
@@ -1088,6 +1088,16 @@ void FGenericPlatformSentrySubsystem::SetContext(const FString& key, const TMap<
 	}
 }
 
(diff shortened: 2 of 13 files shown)


plugin-dev/Source/Sentry/Private/Apple/AppleSentrySubsystem.cpp
```diff
@@ -592,6 +592,19 @@ void FAppleSentrySubsystem::SetContext(const FString& key, const TMap<FString, F
 	}];
 }
 
+void FAppleSentrySubsystem::UpdateContext(const FString& key, const TMap<FString, FSentryVariant>& values)
+{
+	[SENTRY_APPLE_CLASS(SentryObjCSDK) configureScope:^(SentryObjCScope* scope) {
```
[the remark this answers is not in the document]
tustanivsky: The key is correct - https://github.com/getsentry/sentry-cocoa/blob/c34c99b97b112c0da21216f702d4da336c1b2a90/Sources/Sentry/SentryScope.m#L571

plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentrySubsystem.cpp
```diff
@@ -1088,6 +1088,16 @@ void FGenericPlatformSentrySubsystem::SetContext(const FString& key, const TMap<
 	}
 }
 
+void FGenericPlatformSentrySubsystem::UpdateContext(const FString& key, const TMap<FString, FSentryVariant>& values)
+{
+	sentry_update_context(TCHAR_TO_UTF8(*key), FGenericPlatformSentryConverters::VariantMapToNative(values));
+
+	if (crashReporter)
+	{
+		crashReporter->SetContext(key, values);
```
jpnurmi: Is `SetContext` ok here, or should the crash reporter also have `UpdateContext`?

jpnurmi (approved):

plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentrySubsystem.cpp
tustanivsky: `SetContext` worked here since we don't pre-populate any context in the crash-reporter wrapper and thus there's nothing to merge-preserve in the current one-shot `device` case. But for the sake of consistency let's add `UpdateContext` to the crash reporter wrapper as well (https://github.com/getsentry/sentry-unreal/pull/1542/commits/4505fc083be0fa02f89e8101b9e1d4d1c1a2183a).

PR #1548, opened 2026-08-26
Set scope tags in a single call on native platforms

tustanivsky:
When an event is captured with a scope, the plugin copies that scope's tags over to the native SDK. This PR makes it do so in a single `sentry_scope_set_tags` call instead of one `sentry_scope_set_tag` per entry.

## Related items

- https://github.com/getsentry/sentry-native/pull/1993

#skip-changelog


limbonaut (approved):

jpnurmi:
Should we also offer `ISentrySubsystem::SetTags`?

tustanivsky:
> Should we also offer `ISentrySubsystem::SetTags`?

Yes, we can actually benefit from having that in place during SDK initialization when we set up Unreal's default tags. I'll create a follow-up PR for that.

jpnurmi (approved):

jpnurmi:
> > Should we also offer `ISentrySubsystem::SetTags`?
> 
> Yes, we can actually benefit from having that in place during SDK initialization when we set up Unreal's default tags. I'll create a follow-up PR for that.

Btw, I was mainly thinking about offering this to the customer who said something like that they "set 4 tags every frame". I assumed they'd set them via the subsystem API, but I'm not sure...

PR #1558, opened 2026-09-02
Fix eventless attachments being uploaded on Mac/iOS

tustanivsky:
This PR fixes attachments being uploaded on Mac/iOS for events that were never sent, producing attachments not linked to any event.

On Mac/iOS attachments are delivered in a separate envelope once `captureMessage`, `captureEvent` or `captureEnsure` returns. `sentry-cocoa` reports an empty ID for events discarded by the sample rate, a `beforeSend` handler or an event processor, but the upload was performed regardless and capturing an envelope directly bypasses filtering altogether.

Key changes:

- `ISentryId::IsValid` added and implemented for every backend using the empty ID representation of the corresponding SDK: `sentry_uuid_is_nil` (native), `SentryObjCId.empty` (cocoa) and `SentryId.EMPTY_ID` via a new `SentryBridgeJava.isEmptyId` helper (Java)
- `FAppleSentrySubsystem::UploadAttachmentForEvent` skips the upload for events that weren't sent, still removing the file if it was meant to be deleted afterwards

Windows/Linux/Android are unaffected: there attachments belong to the event envelope (or its hint) and are discarded together with the event.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 - Add thread stackwalk mode setting for the native backend ([#1546](https://github.com/getsentry/sentry-unreal/pull/1546))
 
+### Fixes
+
+- Fix eventless attachments being uploaded on Mac/iOS ([#1558](https://github.com/getsentry/sentry-unreal/pull/1558))
+
 ### Dependencies
 
 - Bump Android Gradle Plugin from v6.19.0 to v6.20.0 ([#1553](https://github.com/getsentry/sentry-unreal/pull/1553))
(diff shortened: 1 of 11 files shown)


jpnurmi (approved):

JoshuaMoelans (approved):
LGTM! Could we add a test for this too?

mujacica (approved):

tustanivsky:
> Could we add a test for this too?

Yes, it’d probably make sense to cover filtering-related logic with some tests in a follow-up PR.

PR #1565, opened 2026-09-04
chore: Add iOS support to CI pipeline

tustanivsky:
This PR adds iOS to the CI pipeline. The sample project is built, signed and packaged, then exercised end-to-end on physical devices in the SauceLabs Real Device Cloud mirroring the existing Android setup.

### Key changes

**Code signing**
- `fastlane/` + `match` pulls the development certificate and provisioning profile from `getsentry/codesigning`; requires `MATCH_GIT_PRIVATE_KEY`, `MATCH_PASSWORD` and `APPLE_TEAM_ID` secrets
- Signing config is generated during the build by `scripts/configure-ios-signing.ps1` rather than committed, so local builds keep whatever is configured in the editor
- fastlane is installed via Homebrew since `ruby/setup-ruby`'s prebuilt macOS rubies are hardcoded to the GitHub-hosted runner layout and can't install on Bitrise VM

**Build (`test-ios.yml`)**
- `BuildCookRun` with DDC caching; builds the `.ipa` from the archived `.app` (on a Mac host UE only produces a `.app`) and fails if it isn't signed and provisioned
- Stages the match profile to a fixed path since Unreal reads `IOSProvisioningProfile` as an `FFilePath` and a bare profile name fails config import during the cook

**Integration tests (`integration-test-ios.yml`, `Integration.iOS.Tests.ps1`)**
- Runs on SauceLabs via app-runner's `iOSSauceLabs` provider; covers crash, message, feedback, structured logging, metrics and tracing
- All actions run upfront in one device session so SauceLabs doesn't time out while the harness polls the Sentry API
- App output is read from the log file in the app's Documents container rather than device syslog, which drops UE output under load

**Sample project**
- Enables `bSupportsITunesFileSharing` so the log can be pulled off the device
- `FSentryCrashTest` waits 5s before crashing on mobile: the app died before WebDriverAgent finished verifying the launch, tearing down the Appium session and every remaining action

### Related items

- #1146
- #1463
- #1561
- getsentry/app-runner#32

#skip-changelog


fastlane/Fastfile
```diff
@@ -0,0 +1,18 @@
...
+platform :ios do
+
+  desc "Import signing certificates and provisioning profiles to keychain"
+  lane :prepare_signing do
+    setup_ci if is_ci
+
+    sync_code_signing(
+      type: "development",
+      readonly: true,
+      app_identifier: [CredentialsManager::AppfileConfig.try_fetch_value(:app_identifier)],
```
philprime: `l`: It might be possible to omit this line as fastlane usually falls back to the Appfile internally. Not sure though.

sample/Source/SentryPlayground/IntegrationTests/SentryCrashTest.cpp
```diff
@@ -44,8 +44,8 @@ void FSentryCrashTest::Run()
 
 	Subsystem->SetTag(TEXT("test.crash_id"), EventId);
 
-#if PLATFORM_ANDROID
-	FPlatformProcess::Sleep(1.0f);
+#if PLATFORM_ANDROID || PLATFORM_IOS
+	FPlatformProcess::Sleep(5.0f);
```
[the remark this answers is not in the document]
tustanivsky: Since we already need a delay for the tests to work reliably on SauceLabs Android/iOS devices a few extra seconds is a small trade-off for keeping the code simpler.

philprime (approved):
LGTM

.github/workflows/integration-test-ios.yml
```diff
@@ -0,0 +1,60 @@
...
+
+jobs:
+  integration-test:
+    name: Integration Test
+    runs-on: ubuntu-latest
+
+    env:
+      GITHUB_TOKEN: ${{ github.token }}
+      SAUCE_USERNAME: ${{ secrets.SAUCE_USERNAME }}
+      SAUCE_ACCESS_KEY: ${{ secrets.SAUCE_ACCESS_KEY }}
```
limbonaut, line 25: Should we scope these to the specific step (following the principle of least privilege)?

.github/workflows/integration-test-ios.yml
```diff
@@ -0,0 +1,60 @@
...
+      SENTRY_API_TOKEN:
+        required: true
+
+jobs:
+  integration-test:
+    name: Integration Test
+    runs-on: ubuntu-latest
+
+    env:
+      GITHUB_TOKEN: ${{ github.token }}
```
limbonaut, line 23: This looks unused. Should it be removed?

integration-test/Integration.iOS.Tests.ps1
```diff
@@ -0,0 +1,810 @@
...
+            "$script:SentrySettings`:TracesSampler=/Script/SentryPlayground.CppTraceSampler"
+        )
+
+        Write-Host "Tracing test exit code: $($global:iOSTracingResult.ExitCode)" -ForegroundColor Cyan
+    }
+
+    AfterAll {
+        # Disconnect from iOS device
+        Write-Host "Disconnecting from $Platform..." -ForegroundColor Yellow
+        Disconnect-Device
```
limbonaut, line 209: This keeps the device idle during Sentry polling. Please release it as soon as it's no longer needed.

limbonaut (commented):
Post-merge feedback incoming! I spotted some issues, just leaving here for your consideration @tustanivsky

Issue #476, opened 2024-01-09
Latest Sentry github release doesn't work in Mac


mghimirewhitehat:
### Environment

How do you use Sentry?
Sentry SaaS (sentry.io)

Which version of the SDK?
0.15.1

How did you install the package? (Git-URL, Assetstore)
Git

Which version of Unreal?
5.2.1

Is this happening in Unreal (editor) or on a player like Android, iOS, Windows?
Mac

### Steps to Reproduce

1. Downloaded github package from release page https://github.com/getsentry/sentry-unreal/releases for unreal 5.2
2. Created folder Plugins/Sentry inside the Unreal project
3. When i opened the unreal project it is saying that  "sentry.dylib" cannot be opened because the developer cannot be verified.

### Expected Result

Sentry should run smoothly as expected.

### Actual Result

Sentry is not initialized.

tustanivsky:
@mghimirewhitehat Apparently this issue occur due to some system security settings on your Mac. To run the UE project with Sentry plugin installed please try the following:

1. Open `System Settings`
2. Click `Privacy & Security`, scroll down to the `Security` section and leave this window open for now
3. Run the UE project once again
4. After the popup with aforementioned warning regarding `sentry.dylib` appears click `Cancel` and return to the system settings window
5. Now you should see the `Allow Anyway` button there, click it to proceed (password/touchID input could be required to confirm the operation)
6. Click `Cancel` in warning popup once again
7. Click `Open` in another warning popup that appears right after
8. Unreal Editor should be opened and ready to use Sentry plugin

Please let us know whether it helped.

<img width="924" alt="image" src="https://github.com/getsentry/sentry-unreal/assets/3168564/4eacdf24-2b32-4a65-9339-c18df473c555">

mghimirewhitehat:
Hi @tustanivsky ,

Thank you for your response, this approach works in editor. However when i package the game and codesign with my Developer ID Application certificate it is throwing Developer is not verified pop up in my other Mac, before sentry, it was successfully installed without any issue.

tustanivsky:
While this requires additional investigation on our side here are a few other workarounds that could help:

1. Sentry plugin is available at UE Marketplace so downloading it via the Epic Games launcher instead of using the GitHub package might be something that will allow avoiding the above issues.
[...]

PR #1566, opened 2026-09-05
chore: Sign and notarize macOS binaries

tustanivsky:
This PR adds Developer ID signing and notarization for the macOS binaries shipped in the plugin package.

`SentryObjC.dylib` and `sentry-crash` were only ad-hoc (linker) signed, with no team identifier. Files downloaded through a browser carry macOS' quarantine flag, so loading the plugin from a GitHub release failed with "cannot be opened because the developer cannot be verified" (#476). CI never hit this because runner downloads aren't quarantined.

Signing happens in the existing packaging job via `rcodesign` for release branches.

Note: A bare Mach-O binary can't be stapled, so the notarization ticket stays with Apple and Gatekeeper resolves it online on first load.

Related items:

-  #476

#skip-changelog

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -205,8 +205,8 @@ jobs:
           curl -fsSL "https://github.com/indygreg/apple-platform-rs/releases/download/apple-codesign/${RCODESIGN_VERSION}/apple-codesign-${RCODESIGN_VERSION}-x86_64-unknown-linux-musl.tar.gz" \
             -o rcodesign.tar.gz
           echo "${RCODESIGN_CHECKSUM}  rcodesign.tar.gz" | sha256sum -c -
-          tar -xz --strip-components=1 -f rcodesign.tar.gz rcodesign
-          sudo mv rcodesign /usr/local/bin/rcodesign
+          tar -xz --strip-components=1 -C /tmp -f rcodesign.tar.gz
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -190,6 +196,49 @@ jobs:
...
+          sudo mv /tmp/rcodesign /usr/local/bin/rcodesign
+          rm rcodesign.tar.gz
+
+      - name: Sign macOS binaries
+        if: env.DO_CODESIGN == '1'
+        env:
+          APPLE_CERT_DATA: ${{ secrets.APPLE_CERT_DATA }}
+          APPLE_CERT_PASSWORD: ${{ secrets.APPLE_CERT_PASSWORD }}
+        run: |
+          echo "$APPLE_CERT_DATA" | base64 --decode > "$APPLE_CERT_PATH"
```
limbonaut, line 222: Could add some checks for clearer errors (in case credentials are missing in the repo):
```suggestion
          if [ -z "$APPLE_CERT_DATA" ]; then
            echo "::error title=Codesign failed::Missing Apple signing certificate (APPLE_CERT_DATA)."
            exit 1
          fi
          echo "$APPLE_CERT_DATA" | base64 --decode > "$APPLE_CERT_PATH"
```
With a similar check at notarization step.

limbonaut (approved):
Looks good!

mujacica (approved):

